PR #93, opened 2021-11-16
Stop the logo from moving when back and settings buttons show/hide

rorystephenson:
The buttons are now in fixed-size containers which maintain their side even if the buttons are not shown.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -27,13 +27,21 @@ function App() {
     <Router>
       <div className="App">
         <Navbar>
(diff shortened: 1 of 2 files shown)


rorystephenson:
Don't know if you guys are doing reviews or not so I just tagged you @bastislack for a review in case.

bastislack:
We haven't reviewed the code of each other until now and just committed the smaller changes to main. What would you suggest with your experience on workflows?
Would it make sense to keep the small changes without reviews and only do it for bigger changes or review everything?

rorystephenson:
> Would it make sense to keep the small changes without reviews and only do it for bigger changes or review everything?

Definitely 👍

bastislack (commented):
Finishing review

PR #117, opened 2021-12-08
UI overhaul

rorystephenson:
Creates two different layouts. One for small mobile devices and one for bigger screens with more horizontal space.


bastislack:
I am not sure about the bottom nav on the phone. Having two rows it takes away quite a bit of space. Can you put it back into one row? Otherwise it looks pretty cool!

rorystephenson:
> I am not sure about the bottom nav on the phone. Having two rows it takes away quite a bit of space. Can you put it back into one row? Otherwise it looks pretty cool!

It should actually be only one row so maybe my changes have increased the button sizes, I'll check that.

Otherwise I'm going to rebase and fix the merge conflicts.

rorystephenson:
OK @bastislack I've rebased and I think I've resolved the issue where the bottom nav took two rows on mobile.

I tried to see if the scroll position is saved when going back from adding a trick and it doesn't seem to work for me but it also doesn't seem to work on highline-freestyle.com so I'm not sure why that is.

bastislack:
The scroll position was not restored after adding a trick or a combo because I only saved it when clicking on a specific trick or combo. I extended the FAB now accordingly in c9a4479.

Issue #115, opened 2021-12-07
Updated stickFrequency of trick not shown in combo on backbutton


bastislack:
If you open a combo and then click on one of the tricks change its frequency and press the backbutton the background color of the trick is not updated in the combo

weberax:
The problem for this is, that once created the tricks in the combo are saved in the combo objects, with their stickfrequency at the time of the combo creation... so changing the stickfrequency later does not affect the tricks in the combo.

To fix this we should save in the combo only the id's of the tricks and then query them later for displaying the combodetails

bastislack:
Are you going to work on this or should I have a look at that?

weberax:
I can do the predefined combos and make a db query for you to query a list of id's.
If you could do the combodetail thing would be great

weberax:
did this in #124 ....

weberax:
@bastislack you opened this issue again?
I thought it should be fixed though #124 

When I try to recreate this everything seems to work

bastislack:
Ah I see. The reason why I opened it is because I tried it in the generator and there the colors are not updated. This is because we save the combo as is in the app state and this does not get updated. Should we try to fix it in the generator as well or would you say we just leave it as it is?

PR #124, opened 2021-12-10
Ax/db overhaul

### LINKED ISSUE #92: Update Tricklist and Database

weberax:
this should be the database structure for the beta release ...
closes #92


src/services/db.js
```diff
@@ -9,136 +10,153 @@ export default class Database {
...
-  };
+  // create or update userTrick
+  saveTrick = (trick) => this.db.userTricks.put(trick);
 
   // delete trick
-  deleteTrick = (id) => {
-    if (id < 1000) console.log("can't delete this trick");
-    return this.db.userTricks.delete(Number(id));
-  };
+  deleteTrick = (id) => this.db.userTricks.put({"id": Number(id), deleted: true});
```
bastislack, line 98: This means the user can not delete predefined tricks?

bastislack (approved):

src/predifinedTricksCombos.js
```diff
@@ -0,0 +1,179 @@
...
+10164,,Triple Back Yoda Roll,Davis Hermes,2021,https://www.instagram.com/p/CUNMWEgAGP4/,,,EXPOSURE,EXPOSURE,11,Three Back Yoda Rolls in one bounce.,
+10165,Holy Grail,Barrel Almighty Flip to Barrel Roll,Davis Hermes,2021,https://youtu.be/UdG4DNI0SNk?t=92,92,97,STAND,STAND,11,Start by doing a Barrel Almighty Flip and go for a second roll by grabbing the line and doing a Barrel Roll.,
+10166,,Darth Vader from Feet,Ian Eisenberg,2021,https://youtu.be/HecUpHXRpDw?t=436,436,452,EXPOSURE,EXPOSURE,11,Start in Expo and roll down until you are under the line. Then do a Darth Vader.,
+10167,,Double Back Almighty Flip,Sebastian Egger,2021,https://youtu.be/HecUpHXRpDw?t=547,547,567,EXPOSURE,EXPOSURE,11,Double Back Almighty Flip,Do the first one laid out and tuck for the second one to get good momentum.; Try to get the hook at your ankle. If the line is near your toes it's hard to pull it around for the second roll.
+10168,,Double Front Almighty Flip,Sebastian Egger,2021,https://youtu.be/HecUpHXRpDw?t=780,780,814,EXPOSURE,EXPOSURE,11,Double Front Almighty Flip.,Stay open for the first one and tuck for the second one to get good momentum.`;
+
```
bastislack, line 174: Should I put the same ids in the google sheet?

bastislack (approved):

src/predifinedTricksCombos.js
bastislack: Sorry for the two approvals!

src/services/db.js
weberax: no, the predefined lists are never modified. All the changes the user makes are noted in the user tricks.
If the user wants to delete a predefined trick, it's just marked as deleted in the user tables

src/predifinedTricksCombos.js
weberax: > Should I put the same ids in the google sheet?

yes that would be awesome

src/services/db.js
bastislack: Ok, I see! Cool!

src/predifinedTricksCombos.js
bastislack: > > Should I put the same ids in the google sheet?
> 
> yes that would be awesome

Done!

src/services/db.js
```diff
@@ -9,136 +10,153 @@ export default class Database {
...
-      combos: "++id, name, tricks, minDiff, maxDiff, avgDiff, totalDiff, numberOfTricks, establishedBy, yearEstablished, linkToVideo, comments, stickFrequency"
+    this.db.version(2).stores({
+      // to keep track of all the versions
+      versions: "key,version",
+      // this is the table for the "predefined" tricks, the id's will start from 10000 onwards
+      predefinedTricks: "id,alias,technicalName,establishedBy,yearEstablished,linkToVideo,videoStartTime,videoEndTime,startPos,endPos,difficultyLevel,description,tips,stickFrequency",
+      userTricks: "++id,alias,technicalName,establishedBy,yearEstablished,linkToVideo,videoStartTime,videoEndTime,startPos,endPos,difficultyLevel,description,tips,stickFrequency,deleted",
+      predefinedCombos: "id, name, *tricks, minDiff, maxDiff, avgDiff, totalDiff, numberOfTricks, establishedBy, yearEstablished, linkToVideo, comments, stickFrequency",
+      userCombos: "++id, name, *tricks, minDiff, maxDiff, avgDiff, totalDiff, numberOfTricks, establishedBy, yearEstablished, linkToVideo, comments, stickFrequency, deleted"
```
bastislack, line 20: Should we add startPos and endPos on the combos as well?
We could use it e.g. for combining combos in the future.

bastislack: In the combos we can also add videoStartTime and videoEndTime.

weberax: both we can easily add in the future. I would focus now in this pull request more on getting the combos working nicely. see #115

weberax:
most of it is done, just the stats...

weberax:
So, I think this is done (or good enough...)
Everything should work now. The only thing I wanted to do but couldn't figure out how to do nicely was computing the stats automatically for the predefined combos. but this we probably don't need.
@bastislack either you can have another look at the latest commits or just merge it directly, I think I found a good solution

bastislack (approved):

PR #127, opened 2021-12-13
Add ComboBuilder

bastislack:
Let users create their own custom combos.


weberax:
Maybe we should add some visual hint, that you are right now selecting a trick to add to the combo and that your are not on the normal tricklist screen.
We could probably remove the bottom navbar or change some colors or so....

weberax:
nice looks not too bad, good job getting it to work, i've been trying to change some things but struggling with it for hours...

I've done a merge of the develop branch, now what I think we should do:
- use the ComboDetail component to display the combo in the postCombo page
   for this we can use the randomCombo thing, the same way we use it with the generator
   like this we have the combo stats already included... but I'm having a hard time doing that

@bastislack do you want to have a look at this?

bastislack:
Ok, I am working on it!

PR #139, opened 2021-12-21
Attempt to implement diffLevels and stickFreqs in the generator

bastislack:
@weberax Can you have a look at this? I tried to implement the difficulty and stickFrequency feature for the generator. But it's not working and I am not sure about the use of the liveQuery.


bastislack:
Now it should be working!

Issue #153, opened 2021-12-26
implement a search bar


weberax:
for the pandemic open, as a judge I often need to search a trick. When using the app in the browser it works with the browser search bar, but would be nice if you can also search in the app.

We probably only need to search through names and aliases ... or with an advance option also through description and tricks.

skwidz:
^ Put up a PR to add this feature!

bastislack:
Thanks a lot for your contribution!

PR #156, opened 2021-12-27
Add Fuzzy Search for Tricks

skwidz:
### This PR implements a search bar for the tricks page, addresses #153 
The search is powered by [fuse.js. ](https://fusejs.io/). The search results are shown when the searchPattern is more than the empty-string, otherwise the full tricks list is shown to the user. 

The search operates on the `alias` and `technicalName` fields currently, and could possibly use a bit of tuning but the default settings give a nice result for a first pass. 

**Note:** There is a large section of the `package.lock` that got removed when I added the fuse.js package. Not sure why this happened, but it was automatically done by npm. 

### Before: No search
image

### After: Search bar! 
image


weberax (changes requested):
Nice, thanks for your contribution!

Everything looks good, besides the `package-lock.json`. 
It somehow downgraded the `lockfileVersion`, which should not happen. Maybe your version of npm is not up to date. You can try updating it with `npm i -g npm@latest`.

And and one more thing.
I think we should also not sort and not show the level headers, when we search for something. Otherwise the fuzzy search is a bit strange, because it sorts the results automatically in the order of how much it matches the search term, if you then reorder after we get strange locking search results.
Like when I search `holy`, i expect `Holy Grail` to be on the top, but sorting for levels, shows it only at the bottom!

If you could change this I would merge it.

weberax:
And and one more thing.
I think we should also not sort and not show the level headers, when we search for something. Otherwise the fuzzy search is a bit strange, because it sorts the results automatically in the order of how much it matches the search term, if you then reorder after we get strange locking search results.
Like when I search `holy`, i expect `Holy Grail` to be on the top, but sorting for levels, shows it only at the bottom!

Issue #140, opened 2021-12-21
Generator: Choose possible positions which all the tricks have to have


bastislack:
E.g. if the user chooses only expo and sofa the combo will only have tricks with expo and sofa as positions

Issue #119, opened 2021-12-09
Implement comboGenerator avg.Diff option


bastislack:
As a first approach we could just generate some (10 - 100) combos and then check which one is the closest to the avg. diff.

Issue #110, opened 2021-12-05
Extract generator logic to its own file


bastislack:
Otherwise the ComboGenerator components gets unreadable

PR #159, opened 2022-01-02
ComboGenerator refactoring

weberax:
I just started refactoring the ComboGenerator ...

my ideas:
- [x] export combo finding function to an extra file #110
- [x] instead of randomly ordering all the tricks and then find one which fits, first check for possible next tricks and then choose one at random. (this will improve performance and readability)
- [x] implementing a heuristics for "better" selecting next trick, maybe like this we can create combos with a certain average difficulty
- [x] implement the average difficulty thing #119
- [x] try to restructure the react states
- [x] try to follow more the react guidelines, like not trying to use the `document.getElement()` stuff...

and maybe some other issues like #140


weberax:
@bastislack what else should we do with the generator?

bastislack:
This is awesome! In my opinion there is nothing else to do for now other than what you listed in the PR. After this is done we can focus on getting the difficulty checkboxes working again.

weberax:
done and merged!
@bastislack maybe you can give it some tries if everything works like it should

PR #176, opened 2022-01-30
Fix combo slider

rorystephenson:
Previously decreasing the minimum difficulty or increasing the maximum
after creating a combo would not have any effect. This is because when
copying the old array to create a new one with the extra values we were
copying by reference and then modifying the array which actually means
we were also modifying the old difficultyWhiteList. Therefore when we
added values to the whitelist we also added them to the existing
whitelist which messes with react because when it checks if state has
changed it compares the old white list value with the new one and it
concludes that the value didn't change even though it did.

Maybe clearer explained in steps:

1. newWhiteList = oldWhiteList # both variables now share a reference to
the same piece of memory
2. Modify newWhiteList # This actually modifies both
3. Call function to update the oldWhiteList state to newWhiteList.

Then react compares oldWhiteList to newWhiteList and since we modified
both in considers them the same and nothing changed.


bastislack:
Nice, I didn't know that it was copied by reference!

PR #190, opened 2022-02-08
WIP: Rec prereqs

bastislack:
I can not figure out how to get this to work smoothly...


rorystephenson:
@bastislack I've just pushed a commit which fixes the db querying. This was a helpful reference: https://dexie.org/docs/API-Reference#joining. Some other related helpful topics are:

1. Promises
2. Async code (including the await operator)
3. useLiveQuery

I also bumped the version of the predefined tricks so that the recommended tricks were converted to an array.

There's some things to sort out (like scrolling to the top of the page when clicking a recommended trick) but I'll leave that with you.

bastislack:
Thanks so much Rory! I learned a lot from that commit!

PR #198, opened 2022-02-16
Spanish translations

### LINKED ISSUE #160: Spanish Version
There was some feedback that a spanish version would be much appreciated

rorystephenson:
Filling in translations


rorystephenson:
@bastislack I think I've covered all of the translations but there were a lot and it's likely I have made mistakes. Give it a try when you get a chance.

One thing you'll notice is that I changed some existing english words, e.g.:

StickFrequency -> Stick Frequency
ComboList -> Combo List

The main reason is because you can't really follow the same style in spanish as for example ComboList becomes "Lista de Combos" which I think reads weirdly for a spanish speaker if it's ListaDeCombos since de is a preposition and isn't usually capitalised. Secondly I think that style is a bit unusual if you aren't a developer used to seeing camel case.

src/components/combos/ComboDetails.js
```diff
@@ -188,7 +189,7 @@ const ComboDetails = ({ setUserCombo, comboToShow, addTrickToCombo }) => {
 
           {!inGenerator && !inPostCombo && (
             <div className="boostSkill row justify-content-center">
-              <button className={combo.boostSkill ? "col-8 col-lg-3 col-xl-2 btn btn-warning" : "col-8 col-lg-3 col-xl-2 btn btn-primary" } onClick={toggleBoostSkill}>{combo.boostSkill ? "Unboost this combo" : (<><IoRocketSharp/> Boost this combo</>)}</button>
+              <button className={combo.boostSkill ? "col-8 col-lg-3 col-xl-2 btn btn-warning" : "col-8 col-lg-3 col-xl-2 btn btn-primary" } onClick={toggleBoostSkill}>{combo.boostSkill ? "Unboost this combo" : (<><IoRocketSharp/> <Trans id="comboStats.boostCombo">Boost this combo</Trans></>)}</button>
```
bastislack, line 192: Here we still need to wrap "Unboost this combo" in a <Trans> tag

src/components/tricks/TrickDetails.js
```diff
@@ -202,14 +202,14 @@ const TrickDetails = () => {
...
-            <h4>Set your success frequency:</h4>
+            <h4><Trans id="common.setSuccessFrequency" />:</h4>
             <div onChange={selectFreq}>
               <FreqList stickable={trick}/>
             </div>
           </div>
 
           <div className="boostSkill row justify-content-center">
-            <button className={trick.boostSkill ? "col-8 col-lg-3 col-xl-2 btn btn-warning" : "col-8 col-lg-3 col-xl-2 btn btn-primary" } onClick={toggleBoostSkill}>{trick.boostSkill ? "Unboost this trick" : (<><IoRocketSharp/> Boost this trick</>)}</button>
+            <button className={trick.boostSkill ? "col-8 col-lg-3 col-xl-2 btn btn-warning" : "col-8 col-lg-3 col-xl-2 btn btn-primary" } onClick={toggleBoostSkill}>{trick.boostSkill ? "Unboost this trick" : (<><IoRocketSharp/> <Trans id="trickDetails.boostThisTrick">Boost this trick</Trans></>)}</button>
```
bastislack, line 212: Also here we still need to wrap the "Unboost this trick" in a <Trans> tag

src/services/sortingSchemes.js
```diff
@@ -1,60 +1,67 @@
+import React, { Fragment } from "react"
+import { Trans, SelectOrdinal } from '@lingui/macro';
+import * as common from './commonTranslations';
+import { BsArrowUp, BsArrowDown } from 'react-icons/bs';
+
+const comboTricksTitle = (trickCount) => <SelectOrdinal id="sortingCategory.comboTricksTitle" value={trickCount} other="# Trick Combo" />;
+
 export const trickSortingSchemes = [
-  {"name": "Level Upwards",
+  {"name": <Fragment><Trans id={common.level.id} /> <BsArrowUp className="icon-white" /></Fragment>,
```
bastislack, line 9: I like the idea with the arrows!
But on the phone view the background is white and you can't really see the arrows.
We probably need to adapt the className depending on the screen mode.

src/components/tricks/TrickDetails.js
bastislack: Fixed in [334d334](https://github.com/bastislack/highline-freestyle/pull/198/commits/334d3346a6a6506656230c65abcaddba41153219)

src/components/combos/ComboDetails.js
bastislack: Fixed in [334d334](https://github.com/bastislack/highline-freestyle/pull/198/commits/334d3346a6a6506656230c65abcaddba41153219)

bastislack:
Great work!!
I have two questions left:
1. Should we ask a native spanish speaker to check the translations (e.g. David)?
2. Is it also possible to translate the descriptions and tips of tricks/combos?

bastislack:
Regarding my second question, is it even possible to do this with the <Trans> tag approach? Or do we have to implement it ourselves? Like adding a new column in the google sheet and writing the translations in there and then depending on the locale we decide which entry to display?

src/services/sortingSchemes.js
rorystephenson: Nice I've just pushed a fix for that :)

rorystephenson:
Thanks @bastislack 🙂.

> Should we ask a native spanish speaker to check the translations (e.g. David)?
That would be a good idea. I'm easy on whether we ask someone to look at it now or after deploying and then change any wrong translations.

>Is it also possible to translate the descriptions and tips of tricks/combos?
>... is it even possible to do this with the tag approach? Or do we have to implement it ourselves? Like adding a new column in the google sheet and writing the translations in there and then depending on the locale we decide which entry to display?

Spot on, we'd have to add a new column and use the locale to decide which description to show. It would be a fair bit of work to translate all of the descriptions but it would definitely be great for Spanish speakers. For those translations we'd have to get some motivated Spanish people involved! Maybe that would be nice as a next enhancement for the translations?

bastislack:
> It would be a fair bit of work to translate all of the descriptions but it would definitely be great for Spanish speakers.

I think this would pretty much be the most important feature in the spanish version.
I will add the new column on the google sheet this week and just use an online translator.
I asked David if he wants to fix the translations later, let's see if he is motivated.

bastislack:
I translated the trick descriptions and sent them to David. Let's wait for his feedback.

weberax:
> I translated the trick descriptions and sent them to David. Let's wait for his feedback.

@bastislack do we have some news about this?

bastislack:
He didn't check yet, but he said he will do it :D
I reminded him again now. Let's see. Otherwise we just merge it, I would say?

weberax:
> He didn't check yet, but he said he will do it :D I reminded him again now. Let's see. Otherwise we just merge it, I would say?

top, no problem we can wait. I changed it into its own Milestone...

weberax (changes requested):
before merging a0ef883 has to be reverted

bastislack:
> top, no problem we can wait. I changed it into its own Milestone...

23 out of 200+ translations have been checked for now.
I think it might be better, if we just release the translations as they are and add something like an infobox where people can tell us if a translation is not good. We can still let native speakers check them, but it might be better, if we are not dependent on them.

rorystephenson:
> > top, no problem we can wait. I changed it into its own Milestone...
> 
> 23 out of 200+ translations have been checked for now. I think it might be better, if we just release the translations as they are and add something like an infobox where people can tell us if a translation is not good. We can still let native speakers check them, but it might be better, if we are not dependent on them.

I totally agree!

bastislack:
@JulianDietzel I guess this is completely outdated now. Closing.

Issue #174, opened 2022-01-29
Start testing


bastislack:
We should consider to start testing the app. The sooner we start the better.
Here is a website that explains some things to pay attention to when testing websites: https://www.softwaretestinghelp.com/web-application-testing/

PR #199, opened 2022-02-20
Add testing and fix db race condition

rorystephenson:
This PR adds the jest testing framework as well as a couple of simple initial tests to demonstrate how it works.

My goal was to demonstrate how they are written so we can start writing them for more meaningful functionality but in doing so I actually found a race condition in how we populated the predefined tricks/combos.  Previously we did so as the last step of opening the database but due to how it was written there was nothing preventing queries to the database to be run before the data had been populated which means the queries may return no results when they should.

In order to fix this I've used the db.on('ready') function of Dixie (https://dexie.org/docs/Dexie/Dexie.on.ready#use-case-2-block-dbopen-until-done) to make queries wait for the database to be populated. That also required adding a bunch of "return" operators to the combo/trick population code so that the code actually waits for the returned promises to finish. In order to see the problem of how the db was working before this commit try removing the "return" on line 39 of db.js and then run the tests with `npm test`. You will see that tests fail because logging code finishes after the tests finish since we are not waiting on the data to be populated.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -26,6 +26,7 @@
   ],
   "env": {
     "test": {
+      "presets": ["@babel/preset-env", "@babel/preset-react"],
       "plugins": [
         "dynamic-import-node"
       ]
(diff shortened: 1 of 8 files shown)


bastislack:
Awesome Rory! This will also close #174

Issue #204, opened 2022-03-01
Add more sorting options


bastislack:
Add possibility to sort tricks/combos by other attributes (e.g. starting position).

weberax:
We can keep this an discussion for which other sorting options could be useful.
What we have right now

Tricks:
- Difficulty
- Level
- Frequency
- Starting Position
- Ending Position
- Year

Combo:
- Length
- Stick Frequency
- Year
- Max Difficulty

JulianDietzel:
@weberax What is the difference between "Level" and "Difficulty" here? In the current sorting options there are only the two options of "Level Upwards" and "Level Downwards" no options for "Difficulty".

weberax:
Oh, true difficulty and level is the same in this context...

PR #207, opened 2022-03-26
added starting and end position as sorting option

jumlauft:
Addresses issue #204 to add more sorting options for tricks
Let me know if you have more ideas for sorting options

### DIFF
--- a/src/services/sortingSchemes.js
+++ b/src/services/sortingSchemes.js
@@ -27,6 +27,20 @@ export const trickSortingSchemes = [
     "attributeFunc": (a) => a.stickFrequency,
     "showCategory": false,
   },
+  {"name": "Starting Position",
+  "id": 4,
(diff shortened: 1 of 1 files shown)


weberax:
Nice, thanks a lot for your contribution! (sorry for the late reply... 🤹)

More sorting options could be:
- year (newest tricks at the top)
- average difficulty of combos (maybe rounded?)
- max difficulty of combos

src/services/sortingSchemes.js
```diff
@@ -27,6 +27,20 @@ export const trickSortingSchemes = [
     "attributeFunc": (a) => a.stickFrequency,
     "showCategory": false,
   },
+  {"name": "Starting Position",
+  "id": 4,
+  "sortFunc": (a, b) => {return a.startPos.localeCompare(b.startPos, undefined, {numeric: true, sensitivity: 'base'});},
```
weberax, line 32: I don't see the need for the options, the position are defined in an enum and therefore won't vary in spelling or punctuation

Issue #183, opened 2022-01-31
Consecutive tricks checkbox not working


bastislack:
![Screenshot from 2022-01-31 21-04-19](https://user-images.githubusercontent.com/74077179/151864684-585cfc68-f4fc-48cd-aa18-f5cd67f9049c.png)

bastislack:
Even when the checkbox is disabled consecutive tricks are generated.

weberax:
fixed with #213

PR #213, opened 2022-06-11
Fix incorrect ordering of generated combos

JulianDietzel:
Addresses issue #183, concerning the incorrect generation of consecutive tricks in combos, which should not not have any consecutive tricks.

The problem seemed to stem from two places:
1. The `isAnyComboConditionFulfilled` function did not cover all cases quite right and was very hard to untangle. The function was therefore refactored so that there is now less nesting of if-else-statements. The logical errors were fixed in the process.
2. When loading the tricks using `services/db.js` via the `fillComboWithTricks` function, the returned tricks get sorted so that tricks that have the same id (and therefore are the same trick) are placed right after each other. The issue was fixed by removing the filtering and sorting of the result. Please verify if the filtering and sorting did not actually serve a purpose since to my eyes they didn't seem to change the data in any other way than to alter the ordering.

### DIFF
--- a/src/components/generator/generatorFunction.js
+++ b/src/components/generator/generatorFunction.js
@@ -16,19 +16,30 @@ export function findCombo(vars, retries) {
     maxDifficulty
   ] = vars;
 
-  const isAnyComboConditionFulfilled = (index, lastTrick, currentTrick) => {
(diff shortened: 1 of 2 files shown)


JulianDietzel:
Refactored the `fillComboWithTricks` function to look a lot cleaner and shorter, while maintaining the exact functionality from the last commit.

weberax (approved):
sorry for the laaaaaaaaaaaaate reply :)
looks good, will be merged

PR #225, opened 2023-02-24
Refactor generateFunction.js file

JulianDietzel:
### What?
The generatorFunction.js file and its core methode (old name) `findCombo` were refactored. Mainly renames, cleanups and fixed typos. Also a lot of functionality was moved from one big method into smaller methods.

The functionality stayed the same except for one small thing:
The case when a combo is required to have a certain starting position but the allowed tricks contain no trick with that starting position was not handled and just raised an error. Now the user gets alerted and the function returns.

### Why?
To make the code more legible and prepare it for future changes and fixes.

### DIFF
--- a/src/components/generator/ComboGenerator.js
+++ b/src/components/generator/ComboGenerator.js
@@ -5,7 +5,7 @@ import ComboDetails from '../combos/ComboDetails';
 import { stickFrequencies, positions } from '../../services/enums';
 import useLocalStorage from '../hooks/useLocalStorage';
 import computeStats from '../../logic/combos/computeStats';
-import { findCombo } from './generatorFunction';
+import { generateRandomCombo } from './generatorFunction';
 import Slider, { Range } from 'rc-slider';
 import 'rc-slider/assets/index.css';
 
@@ -81,7 +81,22 @@ const ComboGenerator = ({ difficultyRangeMax, randomCombo, setRandomCombo }) => 
     setRandomCombo(null);
 
     // find the combo with the given parameters
-    const randomTricks = findCombo([tricks, positions, numberOfTricks, startFromCheckbox, startFromPosition, allowDuplicates, allowConsecutiveTricks, allowSimilarPositions, allowTransitions, finishToFeet, avgDifficulty, maxDifficulty]);
+    let conditions = {
+      tricks: tricks,
+      positions: positions,
+      numberOfTricks: numberOfTricks,
+      startFromCheckbox: startFromCheckbox,
+      startingPositionIndex: startFromPosition,
+      allowDuplicates: allowDuplicates,
+      allowConsecutiveTricks: allowConsecutiveTricks,
+      allowSimilarPositions: allowSimilarPositions,
+      allowTransitions: allowTransitions,
+      mustFinishToFeet: finishToFeet,
+      avgDifficulty: avgDifficulty,
+      minDifficulty: minDifficulty,
+      maxDifficulty: maxDifficulty
+    }
+    const randomTricks = generateRandomCombo(conditions);
 
     const { minDiff, maxDiff, avgDiff, totalDiff } = computeStats(randomTricks);
 
--- a/src/components/generator/generatorFunction.js
+++ b/src/components/generator/generatorFunction.js
@@ -1,131 +1,172 @@
 import arePositionsSimilar from '../../logic/combos/similarPositions';
 
-export function findCombo(vars, retries) {
-  const [
-    tricks,
-    positions,
-    numberOfTricks,
-    startFromCheckbox,
-    startFromPosition,
-    allowDuplicates,
-    allowConsecutiveTricks,
-    allowSimilarPositions,
-    allowTransitions,
-    finishToFeet,
-    avgDifficulty,
-    maxDifficulty
-  ] = vars;
+/**
+ * Heuristically generates a random combo which adheres to the specified conditions.
+ * @param {*} conditions Conditions that the combo must meet
+ * @param {number} retries Number of times to retry when generating a combo fails
+ */
+export function generateRandomCombo(conditions, retries= 0) {
+  conditions = {
+    tricks: undefined,
+    positions: [],
+    numberOfTricks: 10,
+    startFromCheckbox: false,
+    startingPositionIndex: 0,
+    allowDuplicates: false,
+    allowConsecutiveTricks: false,
+    allowSimilarPositions: true,
+    allowTransitions: false,
+    mustFinishToFeet: true,
+    avgDifficulty: -1,
+    minDifficulty: 0,
+    maxDifficulty: 11,
+    ...conditions
+  };
 
-  /**
-   * Returns true if the current trick is a legal one in regards to the
-   * specified conditions. This does not check for duplicates as it only
-   * looks at one trick and its predecessor in the combo.
-   */
-  const areLocalComboContitionsLegal = (index, lastTrick, currentTrick) => {
-    const isConsecutiveTricksLegal = allowConsecutiveTricks || lastTrick.id !== currentTrick.id;
-
-    const isEndAndStartEqual = currentTrick.startPos === lastTrick.endPos;
-    const isEndAndStartSimilar = arePositionsSimilar(currentTrick.startPos, lastTrick.endPos);
-    const isTransitionLegal = (allowSimilarPositions && isEndAndStartSimilar)
-                              || isEndAndStartEqual
-                              || allowTransitions;
-
-    if (!isConsecutiveTricksLegal || !isTransitionLegal) {
-      return false;
-    }
-
-    // If the last trick should be to feet, return if that is the case.
-    if (index === numberOfTricks - 1 && finishToFeet) {
-      return ["STAND", "EXPOSURE"].includes(currentTrick.endPos);
-    }
-
-    return true;
+  if (retries >= 100) {
+    alert("could not find a combo, try changing the settings");
+    return;
   }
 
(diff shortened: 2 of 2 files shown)


src/components/generator/ComboGenerator.js
```diff
@@ -81,7 +81,7 @@ const ComboGenerator = ({ difficultyRangeMax, randomCombo, setRandomCombo }) =>
     setRandomCombo(null);
 
     // find the combo with the given parameters
```
WaldemarLehner, line 83: rename to create combo

src/components/generator/ComboGenerator.js
```diff
@@ -81,7 +81,7 @@ const ComboGenerator = ({ difficultyRangeMax, randomCombo, setRandomCombo }) =>
...
-    const randomTricks = findCombo([tricks, positions, numberOfTricks, startFromCheckbox, startFromPosition, allowDuplicates, allowConsecutiveTricks, allowSimilarPositions, allowTransitions, finishToFeet, avgDifficulty, maxDifficulty]);
+    const randomTricks = generateRandomCombo([tricks, positions, numberOfTricks, startFromCheckbox, startFromPosition, allowDuplicates, allowConsecutiveTricks, allowSimilarPositions, allowTransitions, finishToFeet, avgDifficulty, maxDifficulty]);
```
WaldemarLehner, line 84: General JS convention is to create a config object instead of a (very) long tuple.

See Function Definition for further details.

src/components/generator/generatorFunction.js
```diff
@@ -1,131 +1,147 @@
 import arePositionsSimilar from '../../logic/combos/similarPositions';
 
-export function findCombo(vars, retries) {
+/**
+ * Heuristically generates a random combo which adheres to the specified conditions.
+ * @param {*} conditions Conditions that the combo must meet
+ * @param {number} retries Number of times to retry when generating a combo fails
+ */
+export function generateRandomCombo(conditions, retries) {
```
WaldemarLehner, line 8: Accepting a long tuple (conditions) is a bad pattern. It would be better to accept a
config object that overrides some sensible defaults.

`generateRandomCombo` could accept an object like this
```js
const config =  {
    tricks: undefined,
    positions: undefined,
    numberOfTricks: 10,
    startFromCheckbox: false,
    // Rename! not a boolean
    // eg startingPositionIndex
    startFromPosition: 0,
    allowDuplicates: false,
    allowConsecutiveTricks: false,
    allowSimilarPositions: true,
    allowTransitions: false,
    // Rename. Does not sound like
    // a boolean
    // Also. name rather confusing
    // shouldFinishToFeet
    finishToFeet: true,
    avgDifficulty: -1,
    maxDifficulty: 11,
    ...userProvidedConfig
}

```

src/components/generator/generatorFunction.js
```diff
@@ -1,131 +1,147 @@
...
-    if (!isConsecutiveTricksLegal || !isTransitionLegal) {
-      return false;
-    }
+    tricks, positions, numberOfTricks,
+    startFromCheckbox, startFromPosition,
+    allowDuplicates, allowConsecutiveTricks, allowSimilarPositions, allowTransitions, finishToFeet,
+    avgDifficulty, maxDifficulty
+  ] = conditions
+
+  if (!retries) retries = 0
```
WaldemarLehner, line 16: Consider adding a new scope here to avoid potential bugs when changing this in the future

```js
if (!retries) {
    retries = 0;
}
``` 
Alternatively, this could be rewritten as follows:
```js
retries ??= 0
```

src/components/generator/generatorFunction.js
```diff
@@ -1,131 +1,147 @@
...
-    if (index === numberOfTricks - 1 && finishToFeet) {
-      return ["STAND", "EXPOSURE"].includes(currentTrick.endPos);
-    }
+  if (numberOfTricks <= 1) {
+    alert("You need more than one trick for a combo!")
+    return
+  }
 
-    return true;
+  let randomTricks = new Array(numberOfTricks)
```
WaldemarLehner, line 27: Is this needed?

Could initializing an empty array `[]`, following by `push`ing onto said array work aswell?

src/components/generator/generatorFunction.js
```diff
@@ -1,131 +1,147 @@
...
+ * This adds a heuristics which chooses tricks depending on how it fits towards the expected avgDifficulty,
+ * but the difficultyLevels are modified with a random value of a Normal Distribution (which is depending on how
+ * extreme the avgDifficulty is), to have some additional randomness and not get always the same combo.
+ * Original Author: weberax
+ */
+function randomTrickOfAvgDifficulty(possibleTricks, maxDifficulty, avgDifficulty, sumOfDifficulties, i) {
+  for (let j = 0; j < possibleTricks.length; j++) {
+    let min = -maxDifficulty + 2 * Math.abs(maxDifficulty / 2 - avgDifficulty)
+    let max = maxDifficulty - 2 * Math.abs(maxDifficulty / 2 - avgDifficulty)
+    possibleTricks[j].randomDiff = possibleTricks[j].difficultyLevel + randomNormalDist(min, max);
```
WaldemarLehner, line 114: I am not sure what exactly this does. But it might make sense to extract the logic inside the loop into a separate function, e.g.
```js
calculateRandomDiffForTrick(possibleTricks[j])
```

Also.. what exactly is Diff in this context?

src/components/generator/generatorFunction.js
```diff
@@ -1,131 +1,147 @@
...
+  rand = rand / 10.0 + 0.5
+  rand = Math.max(0, Math.min(rand, 1))
+
+  return (max - min) * rand + min
 }
+
+/**
+ * Get a random index for a list of the given length. Index is in [0, length-1]
+ */
+function randomIndex(length) {
```
WaldemarLehner, line 145: Looking at the usages, it might make sense to accept an Array and return the value directly.

e.g. 
```js
function getRandomEntryFromArray(array) {
   return  array[Math.floor(array.length * Math.random())]
}

**Review by WaldemarLehner — CHANGES_REQUESTED** [review]

File: src/components/generator/generatorFunction.js (hunk 4 above)
Comment (JulianDietzel) [reply src/components/generator/generatorFunction.js:16]: I'd propose yet another option: Set the default value of `retries` to 0 in the function header.

File: src/components/generator/generatorFunction.js (hunk 3 above)
Comment (JulianDietzel) [reply src/components/generator/generatorFunction.js:8]: Accepted

File: src/components/generator/generatorFunction.js (hunk 5 above)
Comment (JulianDietzel) [reply src/components/generator/generatorFunction.js:27]: Accepted

File: src/components/generator/ComboGenerator.js (hunk 2 above)
Comment (JulianDietzel) [reply src/components/generator/ComboGenerator.js:84]: Accepted

File: src/components/generator/ComboGenerator.js (hunk 1 above)
Comment (JulianDietzel) [reply src/components/generator/ComboGenerator.js:83]: Out of scope

File: src/components/generator/generatorFunction.js (hunk 7 above)
Comment (JulianDietzel) [reply src/components/generator/generatorFunction.js:145]: Accepted

File: src/components/generator/generatorFunction.js (hunk 6 above)
Comment (JulianDietzel) [reply src/components/generator/generatorFunction.js]: Approved

File: src/components/generator/generatorFunction.js [hunk 8]
```diff
@@ -5,57 +5,71 @@ import arePositionsSimilar from '../../logic/combos/similarPositions';
...
 
-  if (numberOfTricks <= 1) {
+  if (conditions.numberOfTricks <= 1) {
     alert("You need more than one trick for a combo!")
     return
   }
 
-  let randomTricks = new Array(numberOfTricks)
-  let availableTricks = [...tricks]
+  let randomTricks = []
```
WaldemarLehner, line 36: nit: this can be a `const`.
As long as you do not overwrite the variable, e.g. 
```js
const randomTricks = [];
randomTricks = ...
```
const can be used, even when pushing / popping from the Array.

This is similar to how you can mutate the properties of a const object, but not the object itself:
```js
const x = {
   a: 0
}
// allowed
x.a = 42
// not allowed
x = {
  a: 42
}
```

src/components/generator/generatorFunction.js
```diff
@@ -67,20 +81,18 @@ export function generateRandomCombo(conditions, retries) {
...
-  const isTransitionLegal = isLegalSimilar || isEndAndStartEqual || allowTransitions
+  const isLegalSimilar = arePositionsSimilar(currentTrick.startPos, lastTrick.endPos) && conditions.allowSimilarPositions
+  const isTransitionLegal = isLegalSimilar || isEndAndStartEqual || conditions.allowTransitions
 
   if (!isConsecutiveTricksLegal || !isTransitionLegal) {
     return false
   }
 
-  if (index === numberOfTricks - 1 && finishToFeet) {
+  if (index === conditions.numberOfTricks - 1 && conditions.mustFinishToFeet) {
```
WaldemarLehner, line 95: nit: Personally, I would invert these conditions so you go from broad / generic conditions down to the more specific ones. But that is just a personal opinion. Feel free to disregard.

WaldemarLehner (approved):
Looks good to me.

I made two comments, but feel free to disregard.

src/components/generator/generatorFunction.js
JulianDietzel: Approved

src/components/generator/generatorFunction.js
JulianDietzel: Agreed, reads nicer the other way around. Changed it.

Issue #45, opened 2021-11-05
Show trick tips as list 


bastislack:
In the csv file the tips are separated by ';'. Separate those tips and show them in a list.

bastislack:
Using papa parse we can use the transform function (as shown here: https://abaganon.com/tutorials/papaparse.html) to convert the entry in the tricklist to an array and then store it as a multiEntry in Dexie.

weberax:
I would say we don't need to store it as an multiEntry, this would just be useful if we would like make a selected query based on this. But because these are only tips, I think this would never happen.

I would suggest to solve this just on the displaying component...

JulianDietzel:
I'd recommend creating a new table for the tricks. By my understanding, this would be the canonically correct way for a relational database schema (unless Dexie is doing something very unusual here).

This would mean creating a new table `TrickTips(TrickId, Tip)` and then, when loading a trick, querying the new `TrickTips` table for all entries with the id of that trick (`TrickId`). 

[Dexie Documentation for joins](https://dexie.org/docs/API-Reference#joining)

PR #233, opened 2023-03-06
Feature show tips as list

JulianDietzel:
Addresses issue #45 by changing the tips from being a single string to an array of strings, each element containing one tip. If there are no tips for the trick, the `tips` field contains an empty array. The tips column was removed from indexing.
The tips are then displayed as a list in the `TrickDetails` page. The `PostTrick` page now allows adding a variable number of tips.

TrickDetails page:
Screenshot from 2023-03-06 12-49-41

PostTrick page:
Screenshot from 2023-03-10 18-15-00

### Rationale
As there were a number of suggested solutions in the comments to the issue I want to offer my reasoning as to why I consider this to be the best solution.

In contrast to [weberax's comment](https://github.com/bastislack/highline-freestyle/issues/45#issuecomment-996118294) suggesting to solve the issue in the front end, I wanted to have the tips separated into individual tips in the database. This to me is much cleaner as the back end should provide clean data while the front end should be concerned with doing its frontending. Also using a semicolon separated string in the database can cause issues when a tip contains a semicolon as a character.

However I do agree with weberax's comment in that the column does not need to be indexed (and probably shouldn't be, even though, at the current state it is in fact indexed). Looking at Dexies documentation a little more, I found a middle way: The tips can be stored as an array inside the database but don't have to be indexed. So now all tips are converted to arrays but the indexing of the column is removed entirely.

@bastislack I tried using the `transform` function of papa parse, I too think that it might be quite elegant, but I ran into multiple issues and the documentation was a little lacking so I ended up not using it.

### DIFF
--- a/src/components/tricks/TrickDetails.js
+++ b/src/components/tricks/TrickDetails.js
@@ -115,6 +115,13 @@ const TrickDetails = () => {
     });
   }
 
+  function TipList(props) {
+    const listItems = props.tips.map(tip =>
+      <li>{tip}</li>
+    )
+    return (<ul>{listItems}</ul>)
+  }
+
   return (
     <div className="trick-details">
       {trick && (
@@ -160,7 +167,7 @@ const TrickDetails = () => {
           {trick.tips &&
             <div>
               <h3>Tips: </h3>
-              <div className="callout">{trick.tips}</div>
+              <TipList tips={trick.tips} />
             </div>
           }
 
--- a/src/components/tricks/TrickDetails.js
+++ b/src/components/tricks/TrickDetails.js
@@ -119,7 +119,7 @@ const TrickDetails = () => {
     const listItems = props.tips.map(tip =>
       <li key={tip}>{tip}</li>
     );
-    return (<ul>{listItems}</ul>);
+    return (<ul className="callout">{listItems}</ul>);
   }
 
   return (
@@ -164,7 +164,7 @@ const TrickDetails = () => {
             </div>
           }
 
-          {trick.tips &&
+          {trick.tips && trick.tips.length > 0 &&
             <div>
               <h3>Tips: </h3>
               <TipList tips={trick.tips} />
--- a/src/components/tricks/TrickDetails.js
+++ b/src/components/tricks/TrickDetails.js
@@ -117,9 +117,9 @@ const TrickDetails = () => {
 
   function TipList(props) {
     const listItems = props.tips.map(tip =>
-      <li>{tip}</li>
-    )
-    return (<ul>{listItems}</ul>)
+      <li key={tip}>{tip}</li>
+    );
+    return (<ul>{listItems}</ul>);
   }
 
   return (
--- a/src/services/db.js
+++ b/src/services/db.js
@@ -51,6 +51,22 @@ export default class Database {
       });
     });
 
+    this.db.version(8).stores({
+      predefinedTricks: "id,alias,technicalName,establishedBy,yearEstablished,linkToVideo,videoStartTime,videoEndTime,startPos,endPos,difficultyLevel,description,stickFrequency,*recommendedPrerequisites,boostSkill",
+      userTricks: "++id,alias,technicalName,establishedBy,yearEstablished,linkToVideo,videoStartTime,videoEndTime,startPos,endPos,difficultyLevel,description,stickFrequency,*recommendedPrerequisites,deleted,boostSkill",
+    }).upgrade(async tx => {
+      await tx.table("predefinedTricks").toCollection().modify(trick => {
+        if (!Array.isArray(trick.tips)) {
+          trick.tips = trick.tips ? trick.tips.split(";") : []
+        }
+      })
+      await tx.table("userTricks").toCollection().modify(trick => {
+        if (!Array.isArray(trick.tips)) {
+          trick.tips = trick.tips ? trick.tips.split(";") : []
+        }
+      })
+    })
+
     this.db.on('ready', () => {
       // count the tricks in the database and populate it if its empty
       return this.db.versions.get("predefinedTricksVersion").then(ret => {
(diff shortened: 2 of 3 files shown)


src/services/db.js
```diff
@@ -82,40 +98,42 @@ export default class Database {
...
+    // this uses the labels of the csv but, also adds an id
+    const header = trickList.shift().concat(["stickFrequency"]);
+
+    const tricks = trickList.map(trick => {
+      // add 0 for stickFrequency
+      trick.push(0);
+      // make key value pairs
+      return Object.assign.apply({},
+          header.map((v, i) => {
             return ({[v]: trick[i]})
```
WaldemarLehner, line 116: can be shortened to
```js
header.map( (v,i) => ({[v]: trick[i]})
```
although I am wondering why you chose this syntax over a simple
```js
const tricksLookup = {};
header.forEach((entry, i) => {
  tricksLookup[v] = trick[i];
}
return tricksLookup;
```
more readable imo.

src/services/db.js
```diff
@@ -82,40 +98,42 @@ export default class Database {
...
 
-      // adds the tricks to the database
-      return this.db.predefinedTricks.bulkPut(tricks).then(() => {
-        console.log("added tricks to database from the csv");
-        this.db.versions.put({"key": "predefinedTricksVersion", "version": predefinedTricksVersion});
-      });
+    // This turns the list of recommendedPrerequisites (which are separated by a ";") into an Array
+    tricks.map(trick => {
+      if (typeof trick.recommendedPrerequisites === "string") {
+        trick.recommendedPrerequisites = trick.recommendedPrerequisites.split(";").map(string => Number(string));
```
WaldemarLehner, line 124: This is a change from the previous code, so feel free to disregard, but I just wanted to throw this into here.

architecture: Does this need to be a semicolon-separated String? Does the underlying DB support Arrays?

Also: `trick.recommendedPrerequisites` of type `string` gets overwritten with `number[]`
Types without TypeScript are already hard – this makes it needlessly complicated. 
Depending on the time of access, this property is either a `string`, a `number` or a `number[]`.
Could an intermediary object be used here?

WaldemarLehner (changes requested):
Requested some changes, also added some comments about architecture, but feel free to disregard those as they have already been present before this PR

src/components/tricks/TrickDetails.js
```diff
@@ -115,6 +115,13 @@ const TrickDetails = () => {
     });
   }
 
+  function TipList(props) {
+    const listItems = props.tips.map(tip =>
+      <li>{tip}</li>
```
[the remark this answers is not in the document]
JulianDietzel: `tip` is a string and therefore does not have an id. Looking at the [documentation](https://reactjs.org/docs/lists-and-keys.html#basic-list-component) again I agree though, that a key needs to be added. Therefore I'd suggest the following:
```html
<li key={tip}>{tip}</li>
```

src/services/db.js
```diff
@@ -82,40 +98,42 @@ export default class Database {
...
+        trick.recommendedPrerequisites = [trick.recommendedPrerequisites];
+      }
     });
+
+    tricks.map(trick => {
+      trick.tips = trick.tips ? trick.tips.split(";") : []
+    })
+
+    // adds the tricks to the database
+    await this.db.predefinedTricks.bulkPut(tricks)
```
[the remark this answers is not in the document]
JulianDietzel: True. This applies to even more lines. Will take a look and try to update them all.

src/services/db.js
JulianDietzel: I didn't actually write this code but just removed tabs. But while I'm at it I'll implement the former suggestion.

src/components/tricks/TrickDetails.js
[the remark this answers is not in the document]
WaldemarLehner: yep. looks good

src/services/db.js
bastislack: I am managing all the tricks in a google sheet and using a semicolon-separated string seemed like an easy option. However, if there is only one recommendedPrerequisite this is read as a number, which leads to the code I implemented here.
I agree that the types here are quite messy.
Do you see a better option of implementing this?

WaldemarLehner: I think the best solution would be to set `trick.recommendedPrerequisites` in one location only and use other intermediary variables for conversion. This way, there is only one place where change happens. Also makes it easier to maintain later on. 

On the point of maintainability I would also strongly suggest to make use of TypeScript.
I helped @JulianDietzel here and there with some code suggestions on the repo, and we found that trying to infer types from usage caused a significant slowdown during writing code — especially in cases where a variable can be of multiple types.

If you are interested, I could look into moving the repo to TypeScript and Vite.

WaldemarLehner: > I am managing all the tricks in a google sheet and using a semicolon-separated string seemed like an easy option

Rollup (and as a result, Vite) has the [dvs loader](https://github.com/rollup/plugins/tree/master/packages/dsv#readme) Plugin which allows you to simply import a CSV File as a JS Module. I would assume Webpack has something similar. 
This only works for compile-time data. If you are fetching that data from the browser from Google Sheet's API, this won't be useful.

JulianDietzel: Talking to @WaldemarLehner we came to the following conclusion:
- Parsing the CSV File to a JS Object at compile time like he recommended would be nice (as then the conversion does not have to happen every time someone enters the website for the first time) but quite out of scope her
- The way it is handled at the moment is actually alright, considering how the data is passed from the Google Doc via a CSV into the db and it does not necessarily need to be changed
- The code of the data cleaning could be written a good bit cleaner. So I did that and pushed the changes as a new commit ([710df3a](https://github.com/bastislack/highline-freestyle/pull/233/commits/710df3a3746e36b7f9297e0f43110b52cccfef97))

bastislack: > I think the best solution would be to set trick.recommendedPrerequisites in one location only and use other intermediary variables for conversion. This way, there is only one place where change happens. Also makes it easier to maintain later on.

That totally makes sense.

> On the point of maintainability I would also strongly suggest to make use of TypeScript.
I helped @JulianDietzel here and there with some code suggestions on the repo, and we found that trying to infer types from usage caused a significant slowdown during writing code — especially in cases where a variable can be of multiple types.

You are right, that could definitely save some headache with runtime errors and development in general. I will open an issue so we can discuss the migration to TypeScript.

Issue #231, opened 2023-03-05
Left Navbar not responsive to screen height


JulianDietzel:
The navbar on the left side gets cut off if the screen is not high enough. The navbar is not scrollable, so all cut off options are unavailable. This can be an issue when using the app on a computer and scaling it to that size or on when using it on tablets of certain sizes.
The issue might get worse with additional sorting options or if links to for example an imprint are added.

![Screenshot from 2023-03-05 12-01-43](https://user-images.githubusercontent.com/33333283/222956559-aea38822-d454-4b6a-bf82-107f65e11b35.png)

JulianDietzel:
I would propose to do one of two things:
1. Make the navbar scrollable when it gets cut off
2. Move the sorting options out of the navbar. Therefore the navbar becomes so short that the issue will not occur under any reasonable circumstances anymore. The sorting options could be moved next to or underneath the search bar (maybe depending on the screen size), as a drop down menu.

I personally lean towards option 2. It should be fairly easy to implement and the result will visually reflect the semantic connection between the search bar and the sorting options.

bastislack:
I like the second option too. I think it would be great to have a 'filter bar' where users can apply different filters and maybe the sorting options would nicely fit in there?

PR #246, opened 2023-03-17
Move sorting options out of the navbar

Dosbodoke:
fix bastislack/highline-freestyle#231

[Link to issue](https://github.com/bastislack/highline-freestyle/issues/231)

Correct me if this way of creating a PR is not correct, never done PR to open-source before.

## What is the purpose of this PR:

Fix the Navbar responsiveness by approaching the proposed solution.

> Move the sorting options out of the navbar. Therefore the navbar becomes so short that the issue will not occur under any reasonable circumstances anymore. The sorting options could be moved next to or underneath the search bar (maybe depending on the screen size), as a drop down menu. 

## What was done to achieve 

- Created a `SearchBar` component that has both text-based filtering and sort filtering
- Since the component `SearchBar` was already created and the sort filtering would need to be implemented in the Combos page, I enhanced it with text-based filtering [(not fully implemented)](#Know limitation)
- Removed all logic of sorting from both `LeftNav` and `TopNav`
- Removed `sortOpt` state from `App.js`, it should now reduce unnecessarily re-renders 

## How to test if it works?

Open the Tricks Page and see if it looks like this:

image

- Use the text search bar and check if behaves as previously.
- Use the new filter dropdown and check if behaves as previously.

Now use  `Ctrl + Shift + M` to simulate a mobile view and check if the responsiveness is good enough.

## Know limitation

I couldn't make the text-based filtering work on `ComboList.js` so I added a comment on line 30: `// TODO: Finish filtering functionality`


src/components/combos/ComboList.js
```diff
@@ -18,11 +22,14 @@ const ComboList = ({ sortOpt, scrollPosition, setScrollPosition }) => {
   });
 
   // combos query with react hooks -- means it refreshes automaticly
-  const combos = useLiveQuery(() => db.getAllCombos().then(c => c.sort(sortingSchemes[sortOpt].sortFunc)), [sortOpt]);
+  const combos = useLiveQuery(() => db.getAllCombos().then(c => c.sort(comboSortingSchemes[sortOpt].sortFunc)), [sortOpt]);
```
JulianDietzel, line 25: Not really your code so feel free to disregard. But while we are at it: It would be more readable to move the functionality into a separate function. Like so:

```js
const combosFromDBSorted = async () => {
  const allCombos = await db.getAllCombos();
  return allCombos.sort(comboSortingSchemes[sortOpt].sortFunc);
}

// combos query with react hooks -- means it refreshes automatically
const combos = useLiveQuery(() => combosFromDBSorted(), [sortOpt]);
```

src/components/combos/ComboList.js
```diff
@@ -18,11 +22,14 @@ const ComboList = ({ sortOpt, scrollPosition, setScrollPosition }) => {
...
   if (!combos || combos.length == 0) {
     return <p>You have no saved combos. For now it is only possible to create a combo using the combo generator, we are working on supporting custom combos.</p>;
   }
-  console.log(combos)
+
+  // TODO: Finish filtering functionality
```
JulianDietzel, line 30: Finish if possible

JulianDietzel (commented):
Great PR :)

I have not worked in this repo all too much but the way you created the PR seems alright. Usually now there is a review process where issues with the code are fixed and it can be discussed how specific things should or should not be done. The after that @weberax @bastislack merge the PR if they think that it is good.

If looked over your code and ran it myself and it looks pretty good to me. I made some comments on stuff I would change. You can either  accept the changes and implement them or viciously fight me in the comment section (seriously though, I might have gotten stuff wrong too, there were reasons for doing something a certain way I just didn't grasp or you might have an even better way to do it - this is what the comments are there for). To add changes simply push the new commits to your `left-navbar-responsiveness` branch and they will automatically appear here in the PR

Dosbodoke:
I agree with Julian's comment, we need to resolve this Task before merging

> Lastly on a general note, I think it would be good to find out how to get the text-based filtering to work on the ComboList before merging. Otherwise we will have a "broken" feature on the develop branch. Maybe someone else has an idea how to do this.

- [ ] Get the text-based filtering to work on the ComboList.

JulianDietzel:
I think I figured out how to get the search to work. In `ComboList.js` where your TODO comment is try using the following three lines (only the first two are changed, the last one is unaltered):

```js
const searchOptions = { keys: ["name"] };
const fuse = new Fuse(combos, searchOptions);
const searchResults = searchPattern ? fuse.search(searchPattern).map(i => i.item) : combos;
```

JulianDietzel:
Right after commenting I corrected myself. See the following line:
```js
const searchOptions = { keys: ["name", "establishedBy"] };
```
You're just too fast for me :D

Dosbodoke:
I trusted you bro :disappointed, will fix that haha.

Dosbodoke:
Did a force push in order to not mess up a lot with the commits

JulianDietzel:
Also feel free to mark my comments as resolved when you implemented them. (I tend to reply with "Implemented" or something along the lines before resolving the comment so it is clear what the result was. I don't know if this is the best practice approach, but it seems somewhat sensible)

Dosbodoke:
I'm used to work with BitBucket and there we create tasks that need to be marked as resolved in order to allow merge, I think that there is something similar here. 

[GitHub - Task list completed PR check](https://github.com/stilliard/github-task-list-completed)

Don't know if it has a native approach for this.

JulianDietzel:
Resolving the comments is the standard GitHub way to handle this. There are also changes that _need_ to be done before the merge can happen, which can be seen in Pull Request 198 at the moment. My comment on not knowing what the best practice way is, was related to whether to write a brief one word response to the request before closing it or not (but in all honesty, who cares?).

JulianDietzel:
Time. This is not a professional project, so it can take a moment for someone to find the time to check the PR. This PR also affects quite a number of files so checking all of them can't simply be done in a few minutes.

bastislack:
Looks good! Thank you so much for your work :)

PR #255, opened 2023-03-31
Make trick and combo list button fonts consistent

### LINKED ISSUE #254: Fonts inconsistent across trick and combo list pages
The button fonts across the trick and combo list pages are completely different and it would look a lot better if they were the same

Screenshot

Screenshot

DavidKaep:
This fixes #254 by removing `btn-outline-success` from the class-name and making the `combo-preview` consistent with the `trick-preview` in the index.css file.

Heres the result on the combo page:

Screenshot

On the trick page (nothing changed here):

Screenshot

Before:

image

### DIFF
--- a/src/components/combos/ComboDetails.js
+++ b/src/components/combos/ComboDetails.js
@@ -102,7 +102,7 @@ const ComboDetails = ({ setUserCombo, comboToShow, addTrickToCombo }) => {
       <>
       <div className={!inPostCombo ? "col-12" : "col-9"} key={"trick" + trick.id}>
         <Link className="link-to-trick " to={`/tricks/${trick.id}`} key={"trick" + trick.id} >
-          <button className="btn trick-preview skillFreq" freq={trick.stickFrequency}>
+          <button className="btn preview-item skillFreq" freq={trick.stickFrequency}>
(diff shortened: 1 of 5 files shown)


JulianDietzel:
Good change, looks much better to my eyes, too. While at it we could even take this a small step further:

I'd suggest merging the `combo-preview` and `trick-preview` classes into one `preview-item` class and reference this class from both the `ComboList` and the `TrickList`. This way the index.css gets shorter, there is less duplicate code and the chance that the two designs will get out of sync again is lowered greatly.

Another change in the vein of "while we are at it" would be to get rid of the `.combo-preview h3` and `trick-preview h3` sections in the index.css. The `.trick-preview h3` section is already not in use and after your changes `.combo-preview h3` becomes unused as well.

bastislack:
@DavidKaep Nice, I had this on my mind already since a while!
@JulianDietzel's suggestions would be a nice upgrade as well, if you're motivated to implement that (otherwise we can just make a new issue for that) :)

DavidKaep:
I made the changes recommended by @JulianDietzel now the h3 headers are gone from the `index.css` and the previews are merged into a singe `preview-item`

Issue #250, opened 2023-03-22
Extract video functionality from TrickDetails and ComboDetails into separate component


bastislack:
There is quite a lot of duplicate code in TrickDetails and ComboDetails regarding the logic of displaying the video.
It would be better to extract that into its own component.

JulianDietzel:
I'll try to do this refactor over the next few days

JulianDietzel:
Implemented in #256.

PR #256, opened 2023-04-01
Create central VideoEmbed component

JulianDietzel:
Addresses #250, creating a centralized component for embedded videos to avoid duplicate code.

Note: There is a change to the webpack config. This change is only a small one for webpack to also recognize ".jsx" files.

For the future it might be easier to use a module like the [react-social-media-embed](https://www.npmjs.com/package/react-social-media-embed) one to embed videos.
For now the refactor already improves the situation enough though.

### DIFF
--- a/src/components/combos/ComboDetails.js
+++ b/src/components/combos/ComboDetails.js
@@ -13,6 +13,7 @@ import DeleteWarning from '../pop-ups/DeleteWarning';
 import computeStats from '../../logic/combos/computeStats';
 
 import Database from "../../services/db";
+import VideoEmbed from "../misc/video/VideoEmbed";
 const db = new Database();
 
 const ComboDetails = ({ setUserCombo, comboToShow, addTrickToCombo }) => {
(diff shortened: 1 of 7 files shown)


bastislack (commented):
I realized that you used .jsx endings on the files, which we haven't done until now in the repo. Do you think it makes sense to rename the relevant (probably most) files to .jsx to be clear what they are? Maybe this is not relevant anymore when we successfully migrated to TypeScript?

bastislack (commented):
Nice, this looks a lot cleaner now! :)

JulianDietzel:
Yes, I think renaming the appropriate files to `.jsx` would be good. It would make it easier to tell Components and just plain JS files apart. Also, making this changes should take less than 15 minutes, taking very little work.

And yes, this does get obsoleted by switching over to TypeScript, but again, as it takes very little time to make this change I think it is worth it for the meantime. Additionally, when switching to TypeScript, `js` files should become `ts` files and `jsx` files should become `tsx` files. So making the distinction now would probably be helpful when switching over to TS as well.

JulianDietzel:
@bastislack What do you think? Should wait with the change of the file endings for the switch over to TS or do it now? Or, more specifically for this context: Should I keep the file endings of this commit the way they are or stick with the current standard and just rename the them to `.js`?

Issue #266, opened 2023-06-29
Difficulty level 'to be determined'


bastislack:

weberax:
Good idea! 
But should we still give them an estimated level, so they don't bunch up at the end of the list? 
I see the problem of having 4-5  tricks like this with maybe a wide spread of difficulties at the end of the list...

What about marking them as "new" (was talked about before somewhere) and adding little text to say that for new tricks the level is not so clear yet.

bastislack:
We updated the difficulties in the google sheet and we don't have that many undetermined tricks because we did estimate the difficulty for tricks that we could give an opinion on. I feel like we should mainly depend on people up-/downvoting tricks (see #257). But I think it's also cool to integrate the community a bit so we don't have to do all the work (e.g. we created a description saying: 'We haven't figured out yet what's going on here. Look at the video and help us by updating the description.'). I like the idea of motivating the users to help us grade new tricks (via #197).

bastislack:
@WaldemarLehner what is the best value to use in the google sheet to indicate a level is 'to be determined'? Maybe -1 could be better than '?' because it's a number, or just a blank field (so have difficulty level as nullable type)?

WaldemarLehner:
How would you feel about "TBD"?
Common abbreviation for To Be Determined

bastislack:
Ok, I thought that could create some type errors when loading this data into a 'number' object? But if we preprocess the data before loading we can also put 'tbd'.

PR #268, opened 2023-08-28
quick Tricklist update

weberax:
I updated the Tricklist from the google sheet.

To keep the update simple, new collums like `isImportant` or `isVariation` in the sheet were skipped.

Also DifficultLevels with `?` were changed to `999` and displayed as `to be determined`, like this the user can still edit the difficulty based on his interest (leaving the `?` or `tbd` breaks the sorting). This is a dirty hack for #266 we should figure out something better.

@bastislack what do you think? Is this ok?


bastislack:
I guess it should be ok for now!

bastislack (approved):

Issue #239, opened 2023-03-13
Merge predefined* and user* tables


JulianDietzel:
The user tricks and the predefined tricks, just as the user combos and the predefined combos are stored in separate tables in the database. Yet, it is made sure in almost every function that adds to those tables, that there are no duplicate ids between the tables. Also, when tricks or combos are fetched from the database,  the predefined and the user tables are almost always if not always merged. This makes the code harder to read, makes functions longer (as always both tables need to be queried and the results need to be merged) and creates the need for functions like the [mergeLists](https://github.com/bastislack/highline-freestyle/blob/b35b5ff5e098e1f1e512950174ed5f2fd12fd147/src/services/db.js#L124) one.

Couldn't the tables just be merged into one, with an additional column `isPredefined`, which just stores a boolean value and specifies whether the trick is a predefined one or not. I'd estimate that this would shorten the db.js file by about 20% and increase readability a lot. Also it would prevent the tables from getting out of sync, if someone were to accidentally only change / update one of the tables but not the other.

I'd be interested if there is perhaps a good reason for the current architecture that I am missing. If not, I'd suggest merging both the `predefinedTricks` and `userTricks` tables and the `predefinedCombos` and `userCombos` tables.

bastislack:
@weberax was talking about a database refactoring. So maybe this is something we can discuss here?

weberax:
I like the idea!

The original idea of having two different tables, was 
1. to have the `predefinedTricks` seperate to update them easily, without tricks getting mixed up (and of course keeping the user data)
2. and always keep the orginal details of a trick, so that when a user changes attributes of this trick, he can reset them, by just deleting the entry in `userTricks`.

But since we decided to have  the _ids_ in the google form fix, __1.__ should not be a problem anymore.
And __2.__ should also not be a problem.

So to summarize what needs to be done (analogous with the combos):
- combine the two tables, overwriting attributes of the `predefinedTricks` table with the one from the `userTricks`
- adapt all the functions
- add a function to reset one or all tricks back to the default values
- modify the `populateTricks` function so, that when a newer `predefinedTricksVersion` is released only the values, which where not modified by the user are updated (suggest in #194)
   maybe this could be done by only allowing the user only to change certain attributes, like 
   - level
   - alias
   - (what else?)

   which will get saved as "userLevel", "userAlias"... and will overwrite the other values.
   And for Tips i suggest to do something like, merging the predefinded tips with the "userTips".

This might also fix #240

[...]

PR #272, opened 2023-09-07
Rewrite database model

WaldemarLehner:
WIP PR for tracking changes

This PR recreates a new "Core" Library which a UI Layer should integrate with.

Here's a quick overview of what's been done in this PR
- Define Low-Level Dexie Schema
- Define Migration Structure
- Define (coarse) Interface for Interacting with Library
- Define "rich" Objects proxying low-level data representation.
- Write logic for parsing Tricks / Combos from YAML
- Write logic for checking Validity of Tricks / Combos
- Write migration logic for archiving / unarchiving Tricks and Combos

This should be done before merging this PR into the rewrite-Branch

- [x] ~~Discuss and implement all needed access functions~~  
*These will be added as needed by the UI layer*
- [ ] Test Proxy-Objects
- [ ] Further Test Archiving / Unarchiving Logic

#239 - Relevant Discussion


src/App.vue
```diff
@@ -1,4 +1,10 @@
 <script setup lang="ts">
+import {onMounted} from "vue"
+
+onMounted( async() => {
+  (await import("./lib/database/official/index")).default()
```
WaldemarLehner, line 9: This should get a comment stating what this does, as its not really apparent from the import (this starts the trick / combo migration process)

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  return issues;
+}
+
+
+
+
+
+export default async function viteGetAllCombos(tricks: z.infer<typeof DbTricksTableZod>[]) {
+  // @ts-expect-error
```
WaldemarLehner: Add a reason as to why a TS Error is expected here

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,157 @@
...
+          issues.push([e.id, `${i+1}. Entry in recommendedPrerequisites references an ID that does not exist (id: ${f[0]}).`])
+        }
+        else if(f[0] == e.id) {
+          issues.push([e.id, `${i+1}. Entry in recommendedPrerequisites references itself (id: ${f[0]}).`])
+        }
+
+      })
+    }
+    if(e.variationOf) {
+      e.variationOf.map( (f,i) => {
```
WaldemarLehner: Misuse of `.map` - result is not used. Use `.foreach` instead

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,157 @@
...
+function findUndefinedReferences(allTricks: z.infer<typeof DbTricksTableZod>[]) {
+  
+  let idToTrickLookup: Record<number, (z.infer<typeof DbTricksTableZod>)[]> = createRecordLookup(allTricks);
+
+  // issues define the ID of the Trick and an issue message
+  const issues: [number, string][] = []
+
+  allTricks.forEach( e => {
+    if(e.recommendedPrerequisites) {
+      e.recommendedPrerequisites.map( (f,i) => {
```
WaldemarLehner: Misuse of .map - result is not used. Use .foreach instead

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+  })
+
+  return issues;
+}
+
+
+
+export default async function viteGetAllTricks() {
+  // @ts-expect-error
```
WaldemarLehner: Add why this expects a TS error

src/data/variationOfCycleChecker.ts
```diff
@@ -0,0 +1,70 @@
...
+   * here to skip doing redundant work. If a path has already been taken, it can be skipped
+   */
+  const rootsIndex: Record<Trick["id"], Trick["id"][]> = {}
+
+  /**
+   * Create a map of child ID -> parent ID for easy lookup
+   */
+  const childIdToParentIdsLookup: Record<Trick["id"], Trick["id"][]> = {}
+
+  tricks.forEach( e => {
```
bastislack: Maybe this `e` could be renamed to `trick`?

src/lib/database/db-schema.drawio.png
WaldemarLehner, line 1: This needs Updating as there are some minor differences to the current implementation

src/lib/database/official/index.ts
```diff
@@ -0,0 +1,26 @@
+import isSyncNeeded from "./isSyncNeeded";
+import syncCombos from "./syncCombos";
+import syncTricks from "./syncTricks";
+
+export default async function runSyncingProcedure() {
```
WaldemarLehner, line 10: Add JS-Doc describing what this function does

src/lib/database/official/isSyncNeeded.ts
```diff
@@ -0,0 +1,20 @@
+export default async function isSyncNeeded() {
+  // @ts-expect-error No .d.ts defined yet
```
WaldemarLehner: Create a .d.ts File for the virtual module

src/lib/database/official/syncCombos.ts
```diff
@@ -0,0 +1,87 @@
...
+
+
+
+/**
+ * This function will sync combos between the official Repo and the local (browser) database.
+ * 
+ * **Note that this is expected to run AFTER syncTricks!**
+ */
+export default async function syncCombos() {
+  // @ts-expect-error no d.ts
```
WaldemarLehner: Create a .d.ts File for the virtual module

src/lib/database/schemas/Version1Schema.ts
```diff
@@ -0,0 +1,98 @@
...
+  "Double Drop Knee",
+  "Inward Drop Knee",
+  "Leash",
+  "Kneehang",
+  "Soup",
+  "Sit",
+  "Rocket",
+  "Frodo",
+  "Buddha",
+  "Yusus",
```
bastislack: Yisus

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,157 @@
...
+export default async function viteGetAllTricks() {
+  // @ts-expect-error 
+  const pattern = join(fileURLToPath(import.meta.url), "..", "tricks", "*.yaml");
+  const allFilePaths = await globby(pattern)
+
+  const parsedYamlFiles = await Promise.allSettled(allFilePaths.map( e => fetchYamlFile(e)))
+  const erroredFiles = parsedYamlFiles.map( (e,i) => e.status === "rejected" ? ({error: e.reason, path: allFilePaths[i]}) : (undefined as never) ).filter( e => e);
+  const goodFiles = parsedYamlFiles.map( e => e.status === "fulfilled" ? e.value : (undefined as never)).filter( e => e)
+
+  if(erroredFiles.length) {
```
bastislack: Is it possible to use `Array.length` as a boolean expression to use in an `if` statement?

WaldemarLehner: It's possible, and this would work as intended, but `if(arr.length > 0)` would be more descriptive. But I don't remember intentionally keeping it like that. Will change :)

But yes.
`!!0` is false, `!!1` is true :D

src/data/tricks/tricks.schema.json
```diff
@@ -0,0 +1 @@
...pe":"string","minLength":1},"tips":{"type":"array","items":{"type":"string","minLength":1}},"dateAddedEpoch":{"type":"integer","minimum":0},"videos":{"type":"array","items":{"type":"object","properties":{"link":{"type":"string","minLength":1,"format":"uri"},"startTime":{"type":"number","minimum":0},"endTime":{"type":"number","minimum":0}},"required":["link"],"additionalProperties":false}},"startPosition":{"type":"string","enum":["Stand","Exposure","Korean","Sofa","Chest","Back","Belly","Shoulder","Hang","Nevermind","Dropknee","Double Drop Knee","Inward Drop Knee","Leash","Kneehang","Soup","Sit","Rocket","Frodo","Buddha","Yusus","Banana"]},"endPosition":{"$ref":"#/definitions/tricks/allOf/0/properties/startPosition"},"difficultyLevel":{"type":"integer","minimum":0},"showInSearchQueries":{"type":"boolean"}},"required":["id","technicalName","dateAddedEpoch","startPosition","endPosition","difficultyLevel","showInSearchQueries"]},{"type":"object","properties":{"recommendedPrerequisites":{"type":"array","items":{"type":"integer","minimum":0}},"variationOf":{"$ref":"#/definitions/tricks/allOf/1/properties/recommendedPrerequisites"}}}]}},"$schema":"http://json-schema.org/draft-07/schema#"}
```
bastislack: Yisus instead of Yusus

src/data/variationOfCycleChecker.ts
```diff
@@ -0,0 +1,70 @@
...
+    const parents = (e.variationOf ?? []).map( e => e[0])
+    childIdToParentIdsLookup[e.id] = parents
+  })
+
+  // A recursive function that looks at the immediate parent nodes 
+  function handleTraversal(currentId: number, currentChain: Trick["id"][]) {
```
bastislack: This function could be extracted from `buildGraph` to keep it short and concise

src/data/tricks/tricks.schema.json
WaldemarLehner: Gets automatically changed on Build if the linked Zod Type gets updated.

src/data/variationOfCycleChecker.ts
WaldemarLehner: I would keep the function as is — it makes use of data outside the `handleTraversal` scope.
image

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,129 @@
...
+import { parse } from "yaml"
+import { DbCombosTableZod, DbTricksTableZod } from "../lib/database/schemas/CurrentVersionSchema"
+import { readFile } from "node:fs/promises"
+import { ZodError, z } from "zod"
+import {globby} from "globby"
+import { join } from "node:path"
+import { fileURLToPath } from "node:url"
+import { YamlComboTableSchemaZod } from "./viteSchemaGenerator"
+
+async function fetchYamlFile(path: string): Promise<z.infer<typeof DbCombosTableZod>> {
```
bastislack: Is there a reason why you specify the return type for this function but not for the others?

WaldemarLehner: hmm… good question. I think I set it during development to get the IDE to stop complaining :D I ill remove it

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,129 @@
...
+function createRecordLookup<T extends z.infer<typeof DbCombosTableZod> | z.infer<typeof DbTricksTableZod>>(allEntities: T[]) {
+  let idToComboLookup: Record<number, T[]> = {};
+
+  allEntities.forEach(databaseEntity => {
+    if (!idToComboLookup[databaseEntity.id]) {
+      idToComboLookup[databaseEntity.id] = [];
+    }
+      idToComboLookup[databaseEntity.id].push(databaseEntity);
+    });
+  return idToComboLookup;
```
bastislack: I am confused a bit with the name `idToComboLookup` and that the input to the function are all official tricks and combos. Can you maybe explain a bit more on what the function does?

WaldemarLehner: I should rename this to `idToEntityLookup`… function also should get some JSDoc…

but in short it creates a Map of Combo of Trick id => the Entity.

The result is then used to check how many Tricks where are with that ID. (used in `findDuplicateKeys` and `findUndefinedTrickReferences`)

WaldemarLehner: Here's the JSDoc I ill be adding

```ts
/**
 * This takes a list of Database Entities and creates a mapping of ID -> List of Objects
 * 
 * This can be used to detect duplicates or to check if an Entity with a given ID exists.
 */
```

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+      idToEntityLookup[databaseEntity.id] = [];
+    }
+      idToEntityLookup[databaseEntity.id].push(databaseEntity);
+    });
+  return idToEntityLookup;
+}
+
+function findDuplicateKeys(allCombos: z.infer<typeof DbCombosTableZod>[]) {
+
+  let idToTrickLookup: Record<number, (z.infer<typeof DbCombosTableZod>)[]> = createRecordLookup(allCombos);
```
bastislack: Should be `idToComboLookup`

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  return Object.entries(idToTrickLookup).filter(([_,trick]) => trick.length > 1).map(([k,v]) => ({
```
bastislack: I guess `trick` should be something like `tricksInCombo`?

bastislack: Also when I hover over the trick I get the following information:
Bildschirmfoto 2023-09-15 um 12 26 52

Why is there a comboStatus?
Maybe `([_,trick(sInCombo)])` is actually `([_,combo])`?

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+function findUndefinedTrickReferences(allCombos: z.infer<typeof DbCombosTableZod>[], allTricks: z.infer<typeof DbTricksTableZod>[]) {
+  
+  let idToTrickLookup: Record<number, (z.infer<typeof DbTricksTableZod>)[]> = createRecordLookup(allTricks);
+
+  // issues define the ID of the Combo and an issue message
+  const issues: [number, string][] = []
+
+  allCombos.forEach( combo => {
+    if(combo.tricks) {
+      combo.tricks.forEach( (f,i) => {
```
bastislack: `f` could be `trickInCombo`?

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+        if(f[1] !== "official") {
```
bastislack: Can we not use dot notation to access fields? E.g. `trickInCombo.stickableStatus`?

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+          issues.push([combo.id, `All tricks must limit themselves to official ones (not the case for ${i+1}. Entry)`])
```
bastislack, line 69: So we assume, that for now we can only provide combos that consist of 'official' tricks? Why is it not the case for the next entry?

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+
+function findUndefinedTrickReferences(allCombos: z.infer<typeof DbCombosTableZod>[], allTricks: z.infer<typeof DbTricksTableZod>[]) {
+  
+  let idToTrickLookup: Record<number, (z.infer<typeof DbTricksTableZod>)[]> = createRecordLookup(allTricks);
+
+  // issues define the ID of the Combo and an issue message
+  const issues: [number, string][] = []
+
+  allCombos.forEach( combo => {
+    if(combo.tricks) {
```
bastislack, line 64: Maybe there should also be an issue, when there is no (or only one) trick in a combo? 'All combos should have at least two tricks'

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+
+
+export default async function viteGetAllCombos(tricks: z.infer<typeof DbTricksTableZod>[]) {
+  // @ts-expect-error Because there are effectively two TS Projects (Vite Plugin Context 
+  //and Vue Webapp Context), TS gets a bit confused and doesn't think import.meta.url is allowed here
+  const pattern = join(fileURLToPath(import.meta.url), "..", "combos", "*.yaml");
+  const allFilePaths = await globby(pattern)
+
+  const parsedYamlFiles = await Promise.allSettled(allFilePaths.map( comboFilePath => fetchYamlFile(comboFilePath)))
+  const erroredFiles = parsedYamlFiles.map( (fileResult,i) => fileResult.status === "rejected" ? ({error: fileResult.reason, path: allFilePaths[i]}) : (undefined as never) ).filter(Boolean);
```
bastislack, line 92: What does `.filter(Boolean)` do?

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+  const goodFiles = parsedYamlFiles.map( fileResult => fileResult.status === "fulfilled" ? fileResult.value : (undefined as never)).filter(Boolean)
+
+  if(erroredFiles.length) {
+
+    const errorMeta = erroredFiles.map(file => {
+      if(!(file.error instanceof ZodError)) {
+        return file.error+""
```
bastislack: Why not `String(file.error)`?

src/data/viteComboFetcher.ts
bastislack: Nevermind, I just realized this is done in `viteGetAllCombos`

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const duplicateIds = findDuplicateKeys(goodFiles)
+  if(duplicateIds.length > 0) {
+    throw {
+      plugin: "vite-plugin-highline-freestyle-data",
+      message: "Some Combos contain the same ID.\n\n"+JSON.stringify(duplicateIds)
+    }
+  }
+
+  const emptyCombos = goodFiles.filter(combo => !combo.tricks || combo.tricks.length === 0).map( combo => ([combo.id, "Combo needs at least 1 trick"] as const))
```
bastislack: We could either extend `emptyCombos` to also handle the case when only one trick is in a combo, or add a second case 'singleTrickCombo' or something like this. Because combos should consist of at least two tricks.

src/data/viteComboFetcher.ts
```diff
@@ -0,0 +1,134 @@
...
+      plugin: "vite-plugin-highline-freestyle-data",
+      message: "Empty Combos detected.\n\n"+emptyCombos.map( ([comboId, errorMessage]) => `Failed Combo: ${comboId}. Issue: ${errorMessage}`).join("\n")
+    }
+  }
+
+  const undefinedReferences = findUndefinedTrickReferences(goodFiles, tricks)
+  if(undefinedReferences.length > 0) {
+    throw {
+      plugin: "vite-plugin-highline-freestyle-data",
+      message: "Reference Resolution failed.\n\n"+undefinedReferences.map( ([comboId, errorMessage]) => `Failed Combo: ${comboId}. Issue: ${errorMessage}`).join("\n")
```
bastislack: Maybe 'Undefined reference found.' (or something similar) instead of 'reference resolution failed.' ?

src/data/viteDataPlugin.ts
```diff
@@ -0,0 +1,50 @@
...
+        return "\0"+id
+      }
+    },
+    async load(id: string): Promise<string | void> {
+      if(id === "\0virtual:highline-freestyle-data") {
+        return await generateVirtualModuleContent()
+      }
+    }
+  }
+}
```
bastislack, line 50: I don't quite understand the return type of this function. You are returning an object that has a field `name` and then two functions? Don't you have to provide the `function` keyword to declare functions?

src/data/viteSchemaGenerator.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const ShortIdArray = z.array(z.number().int().min(0))
+
+export const YamlTrickTableSchemaZod = DbTricksTableZod.omit({
+  trickStatus: true, 
+  recommendedPrerequisites: true, 
+  variationOf: true
+}).and(z.object({
+  recommendedPrerequisites: ShortIdArray.optional(),
+  variationOf: ShortIdArray.optional()
```
bastislack, line 22: It seems like you're replacing the schema definition of the `DbTricksTableZod`. I understand why you omit the trick/comboStatus but not why you're using the `ShortIdArray.xxx`.

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,158 @@
...
+import {join} from "path";
+import {fileURLToPath} from "url"
+import { ZodError, z } from "zod";
+import { DbTricksTableZod } from "../lib/database/schemas/CurrentVersionSchema"
+import {parse} from "yaml"
+import { readFile } from "fs/promises";
+import { YamlTrickTableSchemaZod } from "./viteSchemaGenerator";
+import buildGraph from "./variationOfCycleChecker";
+
+async function fetchYamlFile(path: string): Promise<z.infer<typeof DbTricksTableZod>> {
```
bastislack: you could omit the return type here as well

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,158 @@
...
+  const asObject = YamlTrickTableSchemaZod.parse(parse(await readFile(path, "utf8")))
+  if(asObject.recommendedPrerequisites) {
+    console.log(asObject.recommendedPrerequisites)
```
bastislack: Maybe add more context or delete the log?

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,158 @@
...
+  
+  let idToTrickLookup: Record<number, (z.infer<typeof DbTricksTableZod>)[]> = createRecordLookup(allTricks);
+
+  // issues define the ID of the Trick and an issue message
+  const issues: [number, string][] = []
+
+  allTricks.forEach( trick => {
+    if(trick.recommendedPrerequisites) {
+      trick.recommendedPrerequisites.forEach( (prerequisitePrimaryKey,i) => {
+        if(prerequisitePrimaryKey[1] !== "official") {
```
bastislack, line 75: As in the comboFetcher: Is it possible to use dot notation here?

src/data/viteTrickFetcher.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  const parsedYamlFiles = await Promise.allSettled(allFilePaths.map( trickFilePath => fetchYamlFile(trickFilePath)))
+  const erroredFiles = parsedYamlFiles.map( (trickYamlFile,i) => trickYamlFile.status === "rejected" ? ({error: trickYamlFile.reason, path: allFilePaths[i]}) : (undefined as never) ).filter(Boolean);
+  const goodFiles = parsedYamlFiles.map( trickYamlFile => trickYamlFile.status === "fulfilled" ? trickYamlFile.value : (undefined as never)).filter(Boolean)
+
+  if(erroredFiles.length > 0) {
+
+    const errorMeta = erroredFiles.map(errorResult => {
+      if(!(errorResult.error instanceof ZodError)) {
+        return errorResult.error+""
```
bastislack: As in the comboFetcher: Why not `String(errorResult.error)`?

src/data/viteDataPlugin.ts
WaldemarLehner: Nope :) JS also allows you to use an arrow function (="lambda" function).

Funnily enough they work *slightly* different. It doesnt matter in this case, but I just thought it would be worth mentioning...

`function(){}` creates a new `this` context, while `()=>{}` just uses the parent's context.

What I mean by this:
```js
#!/usr/bin/env node

const a = {
  x: "bla",
  fn: function() {
    console.log("a:",this)
  }
}

const b = {
  x: "ble",
  fn: () => {
    console.log("b:",this)
  }
}

a.fn()
b.fn()
```

Will log
```
a: { x: 'bla', fn: [Function: fn] }
b: undefined
```
For a, `this` is the object itself,
For b its undefined, because the parent scope does not have `this` specified.

src/data/viteSchemaGenerator.ts
WaldemarLehner: Basically, this is what lets you write the following in the YAML
```yaml
id: 1
...
variationOf:
  - 2
  - 3
```
instead of 
```yaml
id: 1
trickStatus: official
...
variationOf:
  - [2, official]
  - [3, official]
```
It is redundant to write "official" everywhere because we only keep official tricks in the YAML files anyways.

WaldemarLehner: Oh… I should probably read the entire comment before replying :'D

I am replacing it with ShortIdArray, which is defined as an int-Array so that we can use the modified Zod-Schema to parse the YAML files and generate the JSON schema so that your IDE can shout at you when the YAML does not meet the requirements :)

src/data/viteTrickFetcher.ts
WaldemarLehner: Nope. `.forEach` iterates over any prerequisites. Do note however, that `trick.recommendedPrerequisites` does not contain another Trick object, but a Tuple defining the Primary Key.

src/data/viteTrickFetcher.ts
WaldemarLehner: Laziness, mostly. 
`String(x)` is longer than `x+""` :D 

Could switch to String. It's however important to not use the constructor (`new String(...)`), see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String#return_value

src/data/viteComboFetcher.ts
WaldemarLehner: `.filter` accepts a function as its argument. If that function returns thruthy, that entry will be found in the returned array. It its falsy, it wont be found there.

You will usually see something like `array.filter( e => e > 42)`. `e => e > 42` is a function declaration. `Boolean` is the Boolean Constructor. If we use the lambda-Notation, it would be equivalent to `array.filter( e => Boolean(e))`. So it turns e into a true or false value. Do note however that, because the result does not have to be `true` or `false`, but only `truthy` or `falsy`, `array.filter(Boolean)` has the exact same effect as `array.filter( e => e)`

src/data/viteComboFetcher.ts
WaldemarLehner: laziness :P can change

src/data/viteComboFetcher.ts
WaldemarLehner: Yeah… that's some code I wrote before I realized that it doesn't really make sense to have any tricks / combos in the Repo that are not offical.

WaldemarLehner: Modified the block abit…
```ts
allCombos.forEach( combo => {
    if(combo.tricks) {
      combo.tricks.forEach( (trickKey,i) => {
        if(trickKey[1] !== "official") {
          // This should never happen, as the `official` status is added by the Build Process.
          // It's still kept here for now in case that ever changes in the future.
          issues.push([combo.id, `All tricks must limit themselves to official ones (not the case for ${i+1}. Entry)`])
        }
        else if(!idToTrickLookup[trickKey[0]]) {
          issues.push([combo.id, `${i+1}. Entry in tricks references an ID that does not exist (id: ${trickKey[0]}).`])
        }
      })
    }
  })
```

src/data/viteComboFetcher.ts
WaldemarLehner: So in Effect it doesnt really matter, but from what I've heard doing a `filter(Boolean)` will tell TypeScript in the future that the result cannot have undefined's and null's.

So this would not throw a type error:
```ts

const arr: (string | undefined)[] = ["asdf", undefined, "bla"]
               //  vvvvv- this will currently emit a TS error
const arrNoUndef: string[] = arr.filter(Boolean)
```

Right now I am hacking together this assertion by casting `undefined` to `never`.

If a function returns `never`, it… uhhh… never returns :D 

This means an infinite loop or that the function will only exit via throwing an Error.

TS however allows you to return, e.g. `string | never`, which is the same as `string`

so something like `arr.map(e => e === undefined ? undefined as never : e)` will tell TS that undefined will never be returned. 

This is of course not true for this map, but if you immediately add a `.filter(Boolean)` afterwards, it *basically* turns correct :'D

src/data/viteComboFetcher.ts
WaldemarLehner: nope. Thats the primary key tuple, not the entire object

src/data/viteComboFetcher.ts
WaldemarLehner: Renamed to trickKey to make it more clear that this is the Primary Key Tuple

src/data/viteComboFetcher.ts
WaldemarLehner: Yup. Good spot!

src/data/viteComboFetcher.ts
WaldemarLehner: Yep, good shout :)

fyi, `Object.entries` turns the entries of an object into a tuple-Array of `[key,value][]` 

so 
```ts
{
  x: 42
  a: "hello world"
}
```
is turned into 
```ts
[
  ["x", 42],
  ["a", "hello world"], 
]
```
there is also `Object.fromEntries` which will take such a tuple and construct an object from it.

src/data/viteComboFetcher.ts
WaldemarLehner: done

src/data/viteComboFetcher.ts
WaldemarLehner: Changed to filter for `combo => !combo.tricks || combo.tricks.length < 2`, also modified the error message.

src/data/viteComboFetcher.ts
WaldemarLehner: Changed to `Undefined trick reference found`

src/data/viteTrickFetcher.ts
WaldemarLehner: done

src/data/viteTrickFetcher.ts
WaldemarLehner: Removed log, added comment stating that this turns the "simple" IDs into the correct Primary Key tuples

src/data/viteTrickFetcher.ts
WaldemarLehner: done

src/data/variationOfCycleChecker.ts
WaldemarLehner: done

src/data/viteComboFetcher.ts
bastislack: Interesting!
Thanks for the clarification :)

src/data/viteDataPlugin.ts
bastislack: So `resolveId` and `load` are lambda functions with names?

What exactly happens with the two functions, are they just callable from whatever imports `viteDataPlugin`? Or are they executed immediately?

bastislack (approved):

src/lib/database/daos/tricks.ts
JulianDietzel, line 1: Why is this file called `tricks.ts` instead of `trick.ts`?

src/lib/database/daos/combo.ts
```diff
@@ -0,0 +1,358 @@
...
+
+/**
+* This is the "rich" object that should be used by the UI Layer.
+* It contains all the data that the frontend would care about, plus functions to modify, fetch adjacent data, etc.
+*/
+export class Combo implements DbObject {
+  
+  #modified = {
+    dataFromDatabase: {
+      alias: false,
```
JulianDietzel, line 15: Should be called "name"

src/lib/database/daos/combo.ts
```diff
@@ -0,0 +1,358 @@
...
+  public set technicalName(val) {
+    if(val === this.dataFromDatabase.technicalName) {
+      return
+    }
+    this.#modified.dataFromDatabase.technicalName = true;
+    this.dataFromDatabase.technicalName = DbCombosTableZod._def.shape().technicalName.parse(val)
+  }
+
+
+  public get alias() {
```
JulianDietzel, line 64: Should be called "name"

src/lib/database/daos/combo.ts
```diff
@@ -0,0 +1,358 @@
...
+    return this.dataFromDatabase.alias
+  }
+
+  public set alias(val) {
```
JulianDietzel, line 68: Should be called "name"

src/lib/database/db-schema.drawio.png
JulianDietzel, line 1: Needs updating to rename the "alias" attribute of a Combo to "name"

src/lib/database/daos/combo.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+
+  //#endregion
+
+
+
+
+  /**
+   * Creates a new Combo Object. This constructor should be treated as "internal" — The UI Layer should make use of the `combosDao` instead!
```
JulianDietzel: "The UI Layer should make use of the `combosDao` instead!"
Instead of what? For creating new combos? Because at the top of this file it is specified in the comment on the class that this class _is_ the one that the UI layer should use.

src/lib/database/daos/tricks.ts
WaldemarLehner: good shout. Should be `trick.ts`

src/lib/database/official/syncTricks.ts
```diff
@@ -0,0 +1,193 @@
...
+  try {
+    await handleVariationOf()
+  }catch(err) {
+    console.error(`[Official Sync] Failed to update variationOf referencing ${trickId}`, err)
+  }
+  try {
+    await handleTrickReferences()
+  }catch(err) {
+    console.error(`[Official Sync] Failed to combos referencing trick ${trickId}`, err)
+  }
```
JulianDietzel, line 110: To me this looks like all the workings of the top level function (`moveTrick`) should be one ACID-Transaction (Atomic, Consistent, Independent, Durable) on the database. The big two ones here are Atomic and Consistent. Both are violated, as it is possible to run into errors with some of the individual database transactions but still execute the others (not atomic, atomic means either all or nothing) which could potentially result in inconsistent database entries.

The solution could be to just try and do all the transactions with the database but only commit the changes one all of them were successful.

But then also I am seeing this function for the first time, removed from most of its context, so I might be completely off in my assessment.

src/data/viteDataPlugin.ts
WaldemarLehner: They are function definitions stored in the object that gets returned.

That object defines a vite plugin; it is consumed by the vite runtime. Whenever vite has an import to deal with, it sends that import to all plugins' `resolveId`. That function can then either return void / undefined, or return a resolved ID, in which case the Plugin takes over handling the import.

same thing happens with `load`. Can return undefined / void, or a string.

src/lib/database/official/syncTricks.ts
JulianDietzel: The commiting of the canges could, if my understanding of Dexie is correct, even be left out of this function completely and be left to the caller of the function, allowing the caller to create larger ACID-transactions.

src/lib/database/official/syncTricks.ts
JulianDietzel, line 1: A lot of functions in this file might be voilating ACID-Transaction principles. See me only long comment on the first function for details. Am I correct in my observations or am I missing something? Is autocommiting perhaps turned off?

src/lib/database/official/syncCombos.ts
JulianDietzel, line 1: See my comments concerning ACID principles on the `syncTricks.ts` file

src/lib/database/schemas/Version1Schema.ts
```diff
@@ -0,0 +1,98 @@
...
+  startPosition: DbPositionZod,
+  endPosition: DbPositionZod,
+  difficultyLevel: z.number().int().min(0),
+  recommendedPrerequisites: z.array(DbReferenceZod).optional(),
+  variationOf: z.array(DbReferenceZod).optional(),
+  showInSearchQueries: z.boolean(),
+})
+
+export const DbCombosTableZod = z.object({
+  ...CommonDefinition,
```
JulianDietzel, line 70: I'm not very familiar with some of the packages and syntax used here but does this infer, that combos have both a "technicalName" and an "alias" in their DB schema? If so

1. Combos should not have a technical name
2. The name of a combo should be called "name" not "alias"

src/lib/database/schemas/Version1Schema.ts
```diff
@@ -0,0 +1,98 @@
...
+ * 
+ * This is a NOOP for this particular Version (because there is no previous version, so you cannot really do a migration)
+ */
+export async function update(_tx: Transaction) {
+  return;
+}
+
+export const schema = {
+  tricks: "[id+trickStatus], technicalName, alias, establishedBy, yearEstablished, startPosition, endPosition, difficultyLevel, description, *recommendedPrerequisites, tips, *variationOf, showInSearchQueries, dateAddedEpoch, videos",
+  combos: "[id+comboStatus], alias, establishedBy, yearEstablished, *tricks, description, tips, dateAddedEpoch, videos",
```
JulianDietzel, line 96: "name" not "alias"

src/data/combos/example1.yaml
```diff
@@ -0,0 +1,10 @@
+id: 1
```
JulianDietzel, line 1: Perhaps more attributes (like for example a description) could be added here if these examples are used for testing. Else just close this comment.

src/data/variationOfCycleChecker.ts
```diff
@@ -0,0 +1,72 @@
+import {DbTricksTableZod} from "../lib/database/schemas/CurrentVersionSchema"
+import { z } from "zod";
+
+type Trick = z.infer<typeof DbTricksTableZod>
+
+/**
+ * Tricks are set up in a way where variationOf defines a "parent" trick they belong to. 
+ * Important note here is that we do not have trees as a trick can have multiple parents.
+ * 
+ * As such, the set of tricks can be described as a set of disjunct directed cycle-free graphs.
```
JulianDietzel, line 10: This might be a nitpick but are we sure that this graph will always be cycle free? We could enforce it without much of a penalty in the usability but I don't think that it is entirely impossible to run into a situation where there is a cyclic connection. Consider the following simplified idea:

Trick A is a parent to trick B because of the general movement
Trick B is a parent to trick C because of the now more difficult general movement
Trick C is a parent to trick A because of a strange hand movement

Therefore, even though unlikely, it is possible to have a cylce in the graph based on valid data.

JulianDietzel: Or, on a much simpler level: There could be two tricks X and Y which are very similar to each other both in movement and skill level and are therefore marked as each others parents.

src/lib/database/db-schema.drawio.png
JulianDietzel, line 1: In this schema combos do not have a difficulty level even though they should. This needs to be added

src/lib/database/daos/combo.ts
JulianDietzel, line 1: I think that combos should have a difficulty level which is currently missing (for good reason, it was not in the schema). If I am right this would need to be added. I sent out a question to @bastislack on Discord in the group chat about this and I think we should wait for his thoughts on the matter.

JulianDietzel (changes requested):
I'm not done reviewing everything but this should be a decent start.

src/lib/database/daos/combo.ts
WaldemarLehner: You need a db object to create a Combo/Trick Object. The ComboDAO and TrickDAO creates these Combo/Trick Objects for you. 

Ye – the instead is referring to the "raw" access of the Dexie Table — should probably make this more clear

src/data/variationOfCycleChecker.ts
WaldemarLehner: That's a discussion that is out of scope of this PR. I would keep it as is for now and create a new PR in case we decide that cycles are fine.

JulianDietzel: Makes sense but we should keep this in mind

bastislack: I feel like for now we can keep it as it is. As soon as we add tricks that create cycles we will notice it and then we can still discuss if it still makes sense

src/lib/database/db-schema.drawio.png
bastislack: Isn't that something that is described by the tricks? We can use the tricks to calculate `avg./max./min. difficulty`.

src/lib/database/daos/combo.ts
JulianDietzel: I was wrong, combos should not have a difficulty level (other than the one which can be calculated from their tricks, but that should not be an attribute)

src/lib/database/db-schema.drawio.png
JulianDietzel: Yes, I was wrong, combos should not have a difficulty level attribute.

src/lib/database/official/syncTricks.ts
WaldemarLehner: Would being Atomic in this case be *really* desirable? 

Instead of having e.g. one Trick's Metadata fail to get moved to the "archived" status you would have a situation where all Tricks would not get synced.

JulianDietzel: Yes! At least parts. I would argue that the whole thing should be one large atomic transaction, which either updates everything properly or does nothing at all and raises an error which gets reported to either the user or, better, us. This is debatable though.

What I think needs to change for sure is that at least parts of the function should become atomic transactions. The main one I just found is the `handleMetadata` function, which makes a `get`, a `put` and a `delete` interaction with the database. All these interactions are directly linked to each other and they should be one atomic transaction so that the database is still in a consistent state after the function finishes. This is because the data in the database could change between any of the steps or an error could occur while deleting, leaving the Metadata-Schema with a bunch of "duplicate" entries that should not exist.

src/lib/database/daos/combo.ts
WaldemarLehner: Will be pursued in #300

src/lib/database/daos/combo.ts
WaldemarLehner: Will be pursued in #300

src/lib/database/schemas/Version1Schema.ts
WaldemarLehner: Will be pursued in #300

src/data/combos/example1.yaml
WaldemarLehner: Ye. I set these examples just up to see if the Archiving and unarchiving Logic works. 
We could expand on the examples here — but in a separate issue

src/lib/database/daos/combo.ts
WaldemarLehner: Will be pursued in #300

src/lib/database/db-schema.drawio.png
WaldemarLehner: Will be pursued in #300

src/lib/database/schemas/Version1Schema.ts
WaldemarLehner: yep. That's whats happening here. Because the object is defined "as const" you can spread the definition and TS does pick it up correctly.

src/lib/database/daos/combo.ts
WaldemarLehner: JSDoc updated

src/lib/database/daos/tricks.ts
WaldemarLehner: done

PR #277, opened 2023-09-16
Add node and npm version info

JulianDietzel:


README.md
```diff
@@ -4,62 +4,73 @@ This is a collection of most of the freestyle tricks, for people to keep a bette
...
 
 Checkout the [good first issue tag](https://github.com/bastislack/highline-freestyle/labels/good%20first%20issue), if you don't know where to start.
 
+
 ### Requirements:
 
-- node
-- npm
+- Node (v20.6.1)
+- npm (v9.8.1)
```
WaldemarLehner, line 44: Any reason this is not the latest version, `v10.1.0`?

JulianDietzel: I chose it because it is the one that comes with Node v20.6.1, so if someone installs Node version 20.6.1 the person should have npm version 9.8.1 installed by default and they will work well together. I got the info from [this StackOverflow answer](https://stackoverflow.com/a/51243675) referencing [this json file](https://nodejs.org/dist/index.json) (check first item of the file). The info also checks out on my machine.

Therefore I considered dropping the version info for npm entirely but thought it would be better to specify the version explicitly instead of vaguely implicitly through the Node version.

WaldemarLehner: Ye. Personally I don't see the benefit in providing a specific npm version. We dont make use of any recent npm version's features — so as long as you did update your npm version in the last 5 yrs or so you should be good :D

WaldemarLehner (approved):

PR #299, opened 2023-09-18
Add PWA requirements

### LINKED ISSUE #278: Add PWA Support
Add [Vite PWA](https://vite-pwa-org.netlify.app/) support to the rewrite

bastislack:
Make the app an installable PWA.
Closes #278.


WaldemarLehner:
Lighthouse PWA Report: 
image

Only thing thats marked as an issue here is that the [Manifest doesn't have a maskable icon](https://developer.chrome.com/docs/lighthouse/pwa/maskable-icon-audit/?utm_source=lighthouse&utm_medium=devtools)

Given that we don't have an icon configured yet, I wouldn't bother with this for now.

public/logo96.png
WaldemarLehner, line 1: Icon is not suitable for small screens. But that's an issue for later ;)

WaldemarLehner (approved):

pnpm-lock.yaml
WaldemarLehner, line 1: Actually… this shouldnt be here. We make use of package-lock.json — having 2 "sources of truth" will just make things break i.e. "it works on my machine" — we either use pnpm, or npm.

WaldemarLehner (changes requested):

pnpm-lock.yaml
bastislack: I read that pnpm is better than npm. Should we switch to using pnpm, or should we just stick to npm?

WaldemarLehner: It's "better" in the sense that it makes use of centralized cache locally and just sym-links to that cache. 

**BUT** it also means that anyone that wants to work on the Repo *has* to use pnpm — which imo is not something we want.

I personally would like to stick with the default node package manager, npm :D

PR #301, opened 2023-10-28
Trick list overview

bastislack:
Create landing page for the app that shows all tricks in the app.

I am happy with the result for now, except that the app crashes when I refresh the page...
@WaldemarLehner Can you take a look?


WaldemarLehner:
I'ill look at #272 tomorrow. Will look at this afterwards if I find the time :)

WaldemarLehner:
Did some changes :)

I removed the onMounted Hook as it is not needed in this use case.
You usually only need it if you need to access the "real" DOM.
A good example would be if your Vue-App needs to initialize some library onto a "real" HTML-Element (e.g. creating a Three.JS HTML-Canvas)

I also removed the Map and replaced it with a Record, which effectively is a plain JS object.

WaldemarLehner:
Here's what this look like now:
image
image

I would highly recommend conditionally changing the column count OR switching to a flex-container with wrapping enabled with a fixed card size

WaldemarLehner (changes requested):
Looks good but for the Styling, see comment above

PR #303, opened 2023-11-11
Rewrite responsive navbar

JulianDietzel:
I created a responsive navbar with the following five routes:
- Tricks: `/tricks`
- Combos: `/combos`
- Glossary: `/glossary`
- Generator (Combo Generator): `/generator`
- More (This is where I'd put the few settings we have, the imprint and a short About section): `/more`

The navbar comes in only a light theme for now as I think that this is what we should design for first and has language support for English, Spanish and German. It is responsive, switching from a bottom navigation to a side bar at the `lg`-breakpoint of Tailwind. I'm going back and forth whether the `xl`-breakpoint should be used. I think the way I did it is the way though let me know if you know any good reasons for using one or the other.

| Mobile | Desktop |
| - | - |
|  | Bildschirmfoto am 2023-11-11 um 18 46 27 |

I also added one npm script which serves the website in a manner so that other devices of the network can access the page. It can be run with `npm run host`.

@bastislack @Dosbodoke  Le me know if you have any feedback on the design.
@WaldemarLehner Can you take a look and make sure I didn't mess up anything with Vue / TailwindCSS.


JulianDietzel:
After talking to @hell3nacht about it, we slightly changed the highlighting of the router items. I also changed the background to white in order to actually see the navbar with the correct matching background color:

| Mobile | Desktop |
| - | - |
|  |  |

WaldemarLehner:
I havent looked at the code yet... Yet I would still do one Suggestion in regardsvto path naming.

We should stick to a REST-like naming scheme as much as possible.

This would mean getting all combos at /combos, getting a specific combo at /combos/$id and so on.

I think that /generate breaks that convention. I would suggest /combos/new or /combos/create or /combos/new/generate if generation should be wildly different from 'manual' creation .



I would also rename /more to /about, as it gives a clearer description of what this page does.

README.md
```diff
@@ -57,6 +57,8 @@ Checkout the [good first issue tag](https://github.com/bastislack/highline-frees
 2. Start the development server
 
    run `npm run dev`
+   
+   (or `npm run host` if you want to expose the website to the local network)
```
WaldemarLehner, line 61: You could also `npm run dev -- --host` to pass the argument to vite.

src/components/navbars/NavbarBottom.vue
```diff
@@ -0,0 +1,46 @@
+<script lang="ts" setup>
+import { Icon } from '@iconify/vue';
+import { useI18n } from 'vue-i18n';
+
+import messages from "../../i18n/index"
```
WaldemarLehner: When I set up the example I made `/i18n/index` have the keys for the Index-Route.
What do you think about storing the Strings under `/i18n/navbars` instead?

src/components/navbars/NavbarBottom.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const i18n = useI18n({
+  messages,
+  useScope: "local"
+})  
+
+const {t} = i18n;
+
+interface NavLink {
+  tKey: string,
```
WaldemarLehner: nit: translationKey? messageKey?

src/layouts/DefaultLayout.vue
```diff
@@ -1,21 +1,23 @@
...
 <template>
-  <div class="flex flex-col min-h-screen dark:bg-gray-950">
-    <Navbar type="fixed" />
-    <div class="h-14"/>
-    <slot />
-    <div v-if="props.flexSpacer ?? true" class="flex-1"/>
-    <Footer/>
+  <div class="flex flex-col min-h-screen bg-white text-stone-900 m-0 p-0">
+    <div class="flex flex-row m-0 p-0">
+      <div class="collapse lg:visible">
```
WaldemarLehner: What are the implications of using `visibility: collapse` over `display: none`?

src/components/navbars/NavbarSide.vue
```diff
@@ -0,0 +1,51 @@
...
+  to: string,
+  icon: string
+}
+
+const links: NavLink[] = [
+  {tKey: "navbar.tricks", to: "/tricks", icon: "ic:baseline-auto-awesome"},
+  {tKey: "navbar.combos", to:"/combos", icon: "ic:baseline-spoke"},
+  {tKey: "navbar.glossary", to: "/glossary", icon: "ic:sharp-menu-book"},
+  {tKey: "navbar.generator", to: "/generator", icon: "ic:sharp-factory"},
+  {tKey: "navbar.more", to:"/more", icon: "ic:baseline-more-horiz"}
```
WaldemarLehner: `/about` would be more fitting

src/components/navbars/NavbarSide.vue
```diff
@@ -0,0 +1,51 @@
...
+  tKey: string,
+  to: string,
+  icon: string
+}
+
+const links: NavLink[] = [
+  {tKey: "navbar.tricks", to: "/tricks", icon: "ic:baseline-auto-awesome"},
+  {tKey: "navbar.combos", to:"/combos", icon: "ic:baseline-spoke"},
+  {tKey: "navbar.glossary", to: "/glossary", icon: "ic:sharp-menu-book"},
+  {tKey: "navbar.generator", to: "/generator", icon: "ic:sharp-factory"},
```
WaldemarLehner: `/combos/new` or `/combos/new/generate` would be more descriptive and would more closely model REST-Like Structures

WaldemarLehner (changes requested):

README.md
JulianDietzel: Would that be better? Serious question, are there any benefits or downsides to either having a `host` script or passing the `--host` flag and not having a script?

src/components/navbars/NavbarBottom.vue
JulianDietzel: Seems like a good idea. What do you think of the following structure (and is it even a good approach to have multiple .ts files and to have them in separate folders):
```
i18n
    index
        index.ts
        index.de.json
        index.en.json
        ...
    navbar
        navbar.ts
        navbar.de.json
        navbar.en.json
        ...
```

src/components/navbars/NavbarBottom.vue
JulianDietzel: Renamed to `translationKey`. Fixed.

src/layouts/DefaultLayout.vue
JulianDietzel: [Relevant docs](https://tailwindcss.com/docs/visibility)

I just used `collapse` as it was the most sensible option out of the visibility options (`visible`, `invisible` and `collapse`) and didn't even consider using `display: none`. Looking at it now, `display: none` might be more suitable as it allows for other components to change their size. I'll try switching over to this.

JulianDietzel: Fixed by switching to `hidden` class which internally sets `display: none` in order to hide navbars. Used `block` as inverse class.
Kept `collapse` and `visible` for the text on the bottom navbar as this way the icon's size can not change (not a real concern, but I think it is more appropriate to use these to options here).

src/components/navbars/NavbarSide.vue
JulianDietzel: Agreed. I think `/combos/new` would not be good, as to me that route seems more appropriate for the page where a user can create (not generate) a custom combo. 
Would `/combos/new/generate` be better in any way than `/combos/generate`? Otherwise I'd go with the `/combos/generate` option.

src/components/navbars/NavbarSide.vue
JulianDietzel: That is what I called it at first. Then I changed it, as there might be settings under that page as well. Though I think you are still correct and at least the route should be called `/about`. I'll still leave the name of the button ("More") the same though.

JulianDietzel: Changed route from `/more` to `/about`, left button name the same ("More").

README.md
WaldemarLehner: oh no. Keep the `host`script :) 
I just figured the script was added you didn't know about this `-- --$params` notation. If you feel like it would be used a lot, feel free to keep it.

I reckon its useful for checking on Mobile Devices.

Feel free to resolve conversation.

src/components/navbars/NavbarBottom.vue
WaldemarLehner: yup. That looks good!

src/components/navbars/NavbarSide.vue
WaldemarLehner: The question is… in what way do we wish to implement generation.

If it is logically a subpage of creating a new combo, e.g. you click on "Create new" and then click a Dice-Button or something like that, I would vouch for `/combos/new/generate`. 

If `/combos/new` and `/combos/generate` can be seen as fundamentally different routes, `/combos/generate` would work. 

To me it seems that generation would be fundamentally tied to creation. But it really depends on the impl.


I was thinking about a Settings-Button on `/combos/new` which would create a popup where you can select your params and a Dice-Button which would reroll for a new Combo.

The "result" would be placed into the List of Tricks which you could also get by manually adding tricks to the combo. 
In that sense, generation would basically be pre-filling the contents on the `/combos/new` page and you wouldnt even need a `/combos/new/generate` 


So yeah… depends on the impl :P

README.md
JulianDietzel: Sweet, thanks for the explanation. Seeing the webpage on Mobile was the exact reason I created the command.

src/components/navbars/NavbarBottom.vue
JulianDietzel: Implemented it this exact way. Grouped the `index`, `navbar` and `notFound` translations into new separate folders.

src/components/navbars/NavbarSide.vue
JulianDietzel: From all that I have seen and heard so far, the Combo Generator will be a very separate feature with its own tab in the navbar. The process currently (and most probably going forwards) is, that you generate a combo and in case you like it a lot you can save it, which is done by then opening the menu for manually creating a combo and automatically filling in the tricks from the generator.

Therefore, if I read your answer correctly, using `/combos/generate` should be alright shouldn't it?

WaldemarLehner: Yup. combos/generate works for me for now. 
I would like some input from the rest of the team about how they envision `generate` to work, but that's out of scope for now :) 

feel free to resolve

JulianDietzel:
Currently blocked by unresolved issue #305 (missing component library)

JulianDietzel:
#305 has been resolved by PR #311 . Now further work is blocked by the unmerged PR #316. The change is needed in order to be able to use the primary color with different brightness levels. Once that is merged there should be no further issues and I can finish the navbar.

JulianDietzel:
Replaced by PR #318.

Issue #300, opened 2023-09-23
Modify DB Schema for tricks and combos


WaldemarLehner:
To not further bloat the #272 PR the correction of the schema should be done in a separate PR.

Combos should have a name (and no `alias` /  `technicalName`)
Tricks should have an `alias` and a `technicalName` (and no `name`)

Technical Hint: 
Do not forget about the drawio-Graphic!

JulianDietzel:
Is this already implemented?

WaldemarLehner:
nope. Feel free to impl :)

JulianDietzel:
Fixed / Updated in PR #304

PR #304, opened 2023-11-15
(Rewrite) Fix DB naming system for combos

JulianDietzel:
Addresses #300 

I changed all occurrences of of combos having a "technicalName" or an "alias" to just having a "name". The "name" attribute is a mandatory one. 

Updated the drawio-schematic accordingly.

Is there already a way to test the Combo implementation in the DB? If not, please just check the code. Thanks :)


src/lib/database/schemas/Version1Schema.ts
```diff
@@ -93,6 +95,6 @@ export async function update(_tx: Transaction) {
 
 export const schema = {
   tricks: "[id+trickStatus], technicalName, alias, establishedBy, yearEstablished, startPosition, endPosition, difficultyLevel, description, *recommendedPrerequisites, tips, *variationOf, showInSearchQueries, dateAddedEpoch, videos",
-  combos: "[id+comboStatus], alias, establishedBy, yearEstablished, *tricks, description, tips, dateAddedEpoch, videos",
+  combos: "[id+comboStatus], name, establishedBy, yearEstablished, *tricks, description, tips, dateAddedEpoch, videos",
```
WaldemarLehner, line 95: technically, this would warrant a DB Version Bump, but given that there is no release version yet, we can ignore that :D

src/lib/database/schemas/Version1Schema.ts
```diff
@@ -42,34 +42,36 @@ const DbVideoZod = z.object({
 })
 
 
-const CommonDefinition = {
```
WaldemarLehner, line 45: why did you get rid of the Common Definition?

WaldemarLehner (approved):
Approved, although im a bit confused as to why you want the common defs gone

src/lib/database/schemas/Version1Schema.ts
JulianDietzel: Realized this after I had already done it: I got the idea that since the naming system will be different between tricks and combos from here on, I'll have to get rid of the common definition. I later realized that I simply could have moved out the namings and kept the common definitions. But at that point I asked myself whether it would make sense to go back and came to the conclusion that since the two things (tricks and combos) are somewhat but not strongly linked semantically having the common definition primarily would serve the purpose of cutting down the LOC. Since it wouldn't save that many lines, I decided against it.
But if you have a good reason why it would be better to keep the common definition, please let me know, I don't feel strongly about this decision and am happy to revert back.

src/lib/database/schemas/Version1Schema.ts
JulianDietzel: That's what I thought as well. We have not put the current version into production yet, so I went with changing the current version.

src/lib/database/schemas/Version1Schema.ts
WaldemarLehner: I'd prefer to keep the "common" part in one definition — simply so that we don't get into a situation where we do a change on the common part, but forget to at it to one of the schemas.

JulianDietzel: Sure, I'll revert it back

WaldemarLehner (approved):

PR #306, opened 2023-11-25
Remove German language from internationalization

JulianDietzel:
It is what the title says. We decided not to implement German internationalization in the foreseeable future so I removed it entirely for now (it would not be a lot of work to add it back if needed).


WaldemarLehner (approved):
lgtm!

PR #307, opened 2023-11-25
Add Eslint, Prettier and Lint-Staged to Rewrite

### LINKED ISSUE #274: Set up Eslint

### LINKED ISSUE #275: Set up Pre-Commit Hooks to lint before commits

### LINKED ISSUE #276: Setup an action to validate the code with ESLint and build the app
Create a Github Action that verifies that the code complies with ESLint rules. 
Said action should also build the app. We can later extend this action to push the result to Github Pages.

WaldemarLehner:
This PR does the following:

- [x] Add Eslint and Prettier to the Project
- [x] Configure Eslint and Prettier
- [x] Resolve any issues with the Ruleset of Eslint
- [x] Resolve any issues with the Ruleset of Prettier
- [x] Add lint-staged and (if needed) husky for pre-commit hooks
- [x] Create a Github Action to automatically invoke eslint and prettier for changed files in PRs.

Closes #274 
Closes #275 
Closes #276


WaldemarLehner:
Ready to review. 

This PR is… uhh… huge. But, most of the changes are automatically done by eslint and prettier. 
I have fixed the issues discovered by eslint *before* doing the `prettier --fix` run. In the Changes Tab you can look at specific commits. I would recommend looking at the commits before the prettier fix first. 


Once that is done, I would recommend filtering just for Dotfiles in the Changes tab. Anything else is basically irrelevant. 

Oh.. one more thing. As part of the prettier --write run I had to replace the `?` in the Trick YAML files. I replaced them with `42` for now.

image

.vscode/extensions.json
JulianDietzel, line 1: A more general question: Do we actually want the `.vscode` folder checked in? I've always tried to keep my code bases free of editor configs. Now checking StackOverflow people seem to be divided on the issue. I still feel though like the folder should actually be added to the `.gitignore` and be untracked. In case you think this is to broad an issue to discuss here, just send me a message and we can either figure it out in private or open an Issue if necessary.

JulianDietzel (approved):
By my limited understanding the changes look good and the styling defaults are set reasonably. From my side feel free to merge.

.vscode/extensions.json
WaldemarLehner: Ye. I see where you're coming from. I made sure to only check in the suggestions file, which will suggest the user to install the relevant extensions.

Settings and configs related just to myself are not checked in.

JulianDietzel: Even tough running the risk that someone else will not be as careful and include personal files with a `git add --all`  is real I think then we just keep it for now

WaldemarLehner: .gitignore to the rescue :P 

It has the following entry:
```
# Editor directories and files
.vscode/*
!.vscode/extensions.json
```

bastislack:
Oh, we can also get rid of the trick yaml files. Because when everything is working the vite plugin you wrote should fetch the tricks from the sheet, right?

WaldemarLehner:
> Oh, we can also get rid of the trick yaml files. Because when everything is working the vite plugin you wrote should fetch the tricks from the sheet, right? — @bastislack 

Nope. The YAML Files are here to stay as they guarantee that a build is reproducible.
We can however create a Process in which tricks from the google sheet are automatically added (in the form of a Pull Request) as YAML files, e.g. with the help of a Github Action that is run every day or so.

Using PRs would also give another verification step.

WaldemarLehner:
I added another commit which adds the `*.schema.json` files to the prettierignore list.
These files are autogenerated and do not need to be readable.

bastislack:
> Nope. The YAML Files are here to stay as they guarantee that a build is reproducible.

Ok, but then I'll have to complete the tricks, because I only copied a limited version of the tricks into the yaml files.

bastislack (approved):

WaldemarLehner:
> > Nope. The YAML Files are here to stay as they guarantee that a build is reproducible.
> 
> Ok, but then I'll have to complete the tricks, because I only copied a limited version of the tricks into the yaml files.

nope. Idea would be to have an Action that fetches the Sheet and creates the YAML files for you

Dosbodoke:
I forgot to mention about one prettier rule that I like to put on my projects to organize the imports

https://github.com/trivago/prettier-plugin-sort-imports

Example:

Before usage
```javascript
import React, {
    FC,
    useEffect,
    useRef,
    ChangeEvent,
    KeyboardEvent,
} from 'react';
import { logger } from '@core/logger';
import { reduce, debounce } from 'lodash';
import { Message } from '../Message';
import { createServer } from '@server/node';
import { Alert } from '@ui/Alert';
import { repeat, filter, add } from '../utils';
import { initializeApp } from '@core/app';
import { Popup } from '@ui/Popup';
import { createConnection } from '@server/database';
```

After Usage
```javascript
import { debounce, reduce } from 'lodash';
import React, {
    ChangeEvent,
    FC,
    KeyboardEvent,
    useEffect,
    useRef,
} from 'react';

import { createConnection } from '@server/database';
import { createServer } from '@server/node';

import { initializeApp } from '@core/app';
import { logger } from '@core/logger';

import { Alert } from '@ui/Alert';
import { Popup } from '@ui/Popup';

import { Message } from '../Message';
import { add, filter, repeat } from '../utils';
```

It works fine with Vue as well.

PR #308, opened 2023-11-26
Search bar

### LINKED ISSUE #281: Trick List - Search (by name)
As a Freestyler, I would like to search for tricks, so that I can access them more quickly.
Search should...
* be always easily accessible - e.g. by having stick at the top and always appear when scrolling up or via an icon
* allow text input
* search tries to match alias or technical name of tricks
* as soon as user has typed something - maybe change to as soon as more than 2 characters are entered
  * trick list disappears
  * only trick cards that match the input are displayed
  * sorting of trick cards according to percentage of matching?
* have possibility to delete search term fast (e.g. by having x at the right of the search bar)

bastislack:
I am trying to implement the SearchBar. While it worked out when putting the SearchBar in `TrickOverview` I didn't really get it to work when trying to put it into the navbar.
@WaldemarLehner Can you tell what basic concepts I am missing to get this to work? Also emitting the search text all the way up and then passing it to TrickOverview looks kind of ugly...

Should close #281.

### DIFF
--- a/src/components/Navbar.vue
+++ b/src/components/Navbar.vue
@@ -4,6 +4,7 @@ import { Icon } from '@iconify/vue';
 import {isDarkMode, setColorScheme, applyColorScheme} from "../util/colorScheme"
 import { useI18n } from 'vue-i18n';
 import { setNewLocale } from '../util/locale';
+import SearchBar from './SearchBar.vue'
 
 interface Props {
   type: "fixed" | "sticky"
@@ -121,6 +122,7 @@ function toggleDarkmode() {
           </RouterLink>
         </div>
         <div class="flex-1"/>
+          <SearchBar @update-search-query="(searchText) => $emit('updateSearchQuery', searchText)"/>
         <div/>
       </nav>
       <div id="mobile-nav-container" :class="`hidden fixed top-0 overflow-hidden left-0 w-full h-full group/wrapper pointer-events-auto ${isMobileNavExpanded ? 'active' : ''}`">
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -1,0 +1,14 @@
+<script lang="ts" setup>
+import { ref } from 'vue'
+import { Icon } from '@iconify/vue';
+
+const searchText = ref('')
+
+</script>
+
+
+<template>
+  <input v-model="searchText" type="text"
+    class="m-2 mt-4 border-2 border-dark-gray bg-light-gray-250 h-10 px-5 rounded-lg text-sm focus:outline-none"
+    placeholder="Search trick" @input="$emit('update-search-text', searchText)">
+</template>
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -1,14 +1,50 @@
 <script lang="ts" setup>
-import { ref } from 'vue'
+import { ref, nextTick } from 'vue'
 import { Icon } from '@iconify/vue';
 
+const emit = defineEmits(['updateSearchQuery'])
+
 const searchText = ref('')
+const isSearchActive = ref(false)
+const searchInputField = ref(null)
+
+function clearSearchText() {
+  searchText.value = ''
+  emit('updateSearchQuery', searchText.value)
+}
+
+async function activateSearch() {
+  isSearchActive.value = true
+  await nextTick()
+  if (searchInputField.value) {
+    (searchInputField.value as HTMLElement).focus()
+  }
+}
+
+function deactivateSearch() {
+  clearSearchText()
+  isSearchActive.value = false
+}
+
+function onBlur() {
+  if (searchText.value.length === 0) {
+    deactivateSearch()
+  }
+}
 
 </script>
 
 
 <template>
-  <input v-model="searchText" type="text"
-    class="m-2 mt-4 border-2 border-dark-gray bg-light-gray-250 h-10 px-5 rounded-lg text-sm focus:outline-none"
-    placeholder="Search trick" @input="$emit('update-search-text', searchText)">
+  <div class="relative flex z-30 bg-white border-b-2 border-dark-gray">
+    <Icon icon="ic:round-search" @click="activateSearch"
+      :class="`absolute z-30 text-lg ${isSearchActive ? 'left-4' : 'right-4'}`" />
+
+    <input v-if="isSearchActive" v-model="searchText" ref="searchInputField" type="text"
+      class="w-full m-2 border-2 border-dark-gray bg-light-gray-250 h-10 z-20 px-7 rounded-lg text-sm focus:outline-secondary"
+      placeholder="Search trick" @input="$emit('updateSearchQuery', searchText)" @blur="onBlur">
+
+    <Icon icon="ic:round-cancel" v-if="searchText" @click="deactivateSearch"
+      class="absolute right-4 z-40 text-lg" />
+  </div>
 </template>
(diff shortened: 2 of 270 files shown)


src/components/SearchBar.vue
```diff
@@ -0,0 +1,50 @@
+<script lang="ts" setup>
+import { ref, nextTick } from 'vue'
+import { Icon } from '@iconify/vue';
+
+const emit = defineEmits(['updateSearchQuery'])
```
WaldemarLehner, line 5: untyped emit. use 
```ts
defineEmits<{ updateSearchQuery: [data: string] }>()
```
See https://vuejs.org/api/sfc-script-setup.html#type-only-props-emit-declarations

WaldemarLehner: image

src/components/SearchBar.vue
```diff
@@ -0,0 +1,50 @@
...
+const searchInputField = ref(null)
+
+function clearSearchText() {
+  searchText.value = ''
+  emit('updateSearchQuery', searchText.value)
+}
+
+async function activateSearch() {
+  isSearchActive.value = true
+  await nextTick()
```
WaldemarLehner, line 18: what does this do?

src/components/Navbar.vue
```diff
@@ -121,6 +122,7 @@ function toggleDarkmode() {
           </RouterLink>
         </div>
         <div class="flex-1"/>
+          <SearchBar @update-search-query="(searchText) => $emit('updateSearchQuery', searchText)"/>
```
WaldemarLehner, line 125: dont use `$emit`. Create a strongly typed Emit definition instead (see previous comment)

src/components/SearchBar.vue
```diff
@@ -0,0 +1,50 @@
...
+
+
+<template>
+  <div class="relative flex z-30 bg-white border-b-2 border-dark-gray">
+    <Icon icon="ic:round-search" @click="activateSearch"
+      :class="`absolute z-30 text-lg ${isSearchActive ? 'left-4' : 'right-4'}`" />
+
+    <input v-if="isSearchActive" v-model="searchText" ref="searchInputField" type="text"
+      class="w-full m-2 border-2 border-dark-gray bg-light-gray-250 h-10 z-20 px-7 rounded-lg text-sm focus:outline-secondary"
+      placeholder="Search trick" @input="$emit('updateSearchQuery', searchText)" @blur="onBlur">
```
WaldemarLehner, line 45: dont use `$emit`. Create a strongly typed Emit definition instead (see previous comment)

src/util/tricks.ts
WaldemarLehner, line 1: Use YAML-Defined Tricks instead

src/util/trick.ts
WaldemarLehner, line 1: use Trick Proxy Object created from the YAML Defined tricks instead

src/routes/IndexRoute.vue
```diff
@@ -1,26 +1,15 @@
...
 
 <template>
-  <DefaultLayout :flex-spacer="false">
-    <div class="flex flex-1 flex-col justify-center  self-stretch ">
-      <div class="flex flex-row justify-center">
-        <h1 class="text-5xl font-bold dark:text-white transition-all duration-200 hover:-translate-y-3">{{ t('greeting') }}</h1>
-      </div>
-    </div>
+  <DefaultLayout :flex-spacer="true" @update-search-query="(searchQuery) => searchText = searchQuery">
+    <TrickOverview :searchText="searchText"/>
```
WaldemarLehner, line 13: This is the correct way to pass events

src/routes/IndexRoute.vue
```diff
@@ -1,26 +1,15 @@
...
+import { ref } from 'vue';
 
-import messages from "../i18n/index"
-
-const i18n = useI18n({
-  messages,
-  useScope: "local"
-})  
-
-const {t} = i18n;
```
WaldemarLehner, line 12: Why was Internalization removed from the Index-Route?

Feel free to close if it was moved into the child component

src/components/TrickOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  tricksByDifficulty[difficultyLevel].push(trick)
+})
+
+Object.keys(tricksByDifficulty).forEach(
+  e => tricksByDifficulty[e as any] = tricksByDifficulty[e as any].sort(compareTrickNames)
+);
+
+console.log(tricksByDifficulty)
```
WaldemarLehner, line 31: remove debug log before merging

src/layouts/DefaultLayout.vue
```diff
@@ -12,8 +12,8 @@ const props = defineProps<Props>()
 
 <template>
   <div class="flex flex-col min-h-screen dark:bg-gray-950">
-    <Navbar type="fixed" />
-    <div class="h-14"/>
+    <Navbar type="fixed" @update-search-query="(searchText) => $emit('updateSearchQuery', searchText)"/>
```
WaldemarLehner, line 15: here as well: $emit -> emit with typing

WaldemarLehner (changes requested):

WaldemarLehner:
Also do run `git merge rewrite` as this branch is quite outdated.

src/components/SearchBar.vue
bastislack: I wanted to give focus to the input field but it didn't work without the nextTick() function...

bastislack: I think it waits until the rebuild is done and then executes the rest of the code

src/routes/IndexRoute.vue
bastislack: Ok, so that means all the variables that are passed to some child components will appear in the `IndexRoute` and then passed as props to whereever they are needed. Aren't we gonna end up with lots of variables in here then?

src/components/TrickOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+
+</script>
+
+
+<template>
+  <div v-if="!isSearchActive">
+    <div v-for="level in Object.keys(tricksByDifficulty).length - 1" :key="level">
+      <hr class="border-dark-gray">
+      <h2 class="text-2xl font-bold py-2 text-center bg-light-gray">
+        Level {{ level }}
```
JulianDietzel, line 75: Internationalization missing

src/components/TrickOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+          class="p-2 border-2 rounded-md border-dark-gray bg-light-gray-250 flex items-center justify-center container h-20 drop-shadow">
+          <p class="text-dark-gray text-center line-clamp-3 text-base/5 break-words">
+            {{ trick.alias && trick.alias.length > 0 ? trick.alias : trick.technicalName }}
+          </p>
+        </div>
+      </div>
+    </div>
+    <hr class="border-dark-gray">
+    <h2 class="text-2xl font-bold py-2 text-center bg-light-gray">
+      To Be Determined
```
JulianDietzel, line 89: Internationalization missing

src/components/TrickOverview.vue
```diff
@@ -0,0 +1,116 @@
...
+        class="p-2 border-2 rounded-md border-dark-gray bg-light-gray-250 flex items-center justify-center container h-20 drop-shadow">
+        <p class="text-dark-gray text-center line-clamp-3 text-base/5 break-words">
+          {{ trick.alias && trick.alias.length > 0 ? trick.alias : trick.technicalName }}
+        </p>
+      </div>
+    </div>
+  </div>
+  <div v-else>
+    <p class="text-center text-2xl font-bold">
+      No tricks found :(
```
JulianDietzel, line 113: Internationalization missing

JulianDietzel (changes requested):
Found some words that need internationalization before merging. Refer to [this wiki page](https://github.com/bastislack/highline-freestyle/wiki/Internationalization) on how to add internationalization to components :)

JulianDietzel:
The issue was addressed and implemented in #351.

Issue #309, opened 2023-11-27
Add a theme (and how to go about it)


JulianDietzel:
Opening this since I have used standard TailwindCSS colors for the navbar in PR #303, @bastislack defined a custom color scheme in PR #308  and (if I am correct) @Dosbodoke updated the tailwind.css file to play well with shadcn in PR #302. It's all valid independently but a mess in combination.

## Defining the approach
What needs to happen first is we need to define how we go about theming on a technical level and what kind of colors we need.

### The technical side
In #305 we agreed on using shadcn as a component libary. Shadcn also comes with a [theming system](https://ui.shadcn.com/themes) that I strongly assume automatically works with the shadcn components. That's really nice and convenient and I think we should therefore consider using it.

### Practical changes
An issue I have with the shadcn colors is that they seem quite limited. The primary and accent color come in one version, the color itself. Often, they are needed in darker and lighter versions for different purposes though. Therefore I would suggest that in addition to the following colors themselves we provide each of them in both three lightened and three darkened versions: 
- primary
- secondary
- accent
- destructive

To stick close to the [TailwindCSS color naming](https://tailwindcss.com/docs/customizing-colors), versions  could then have a naming along the lines of 
- `primary-50`
- `primary-100`
- `primary-300`
- `primary`
- `primary-700`
- `primary-900`
- `primary-950`

The name of `primary` is left as is (instead of the logical `primary-500`) to maintain interoperability with shadcn and to highlight what the main version of the color is supposed to be.

### Adding "skill-level" colors
One of the central elements of the app it to track user progress. This has traditionally been displayed by giving the trick cards different colors depending on how good the user is at that trick. After some consideration I would personally like to stick to that approach. As a result we need either six or seven color for that, depending on whether "Work in progress" should be indicated by one of these colors. Since that is a discussion for a later day (I think) we should design for 7 colors and drop the one if we decide against it.

### To dark theme or not to dark theme (Not)
For simplicity reasons I would ditch having a dark theme option. Having the option creates development overhead and I have personally not once used the app and wished for a dark theme. We could keep the logic in, in case we one day want to have a dark version but for now I wouldn't bother with it.

### What we would need as a result:

Most of these colors could be taken from one of the provided themes of shadcn.

| name | 50 | 100 | 300 | standard | 700 | 900 | 950 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| background | | | | :heavy_check_mark:  | | | |
[...]

PR #311, opened 2023-11-29
Add Shadcn

Dosbodoke:
### What is the purpose of this PR?
Add [shadcn](https://www.shadcn-vue.com/docs/installation/vite.html) as the project component library.

### What was done to achieve it?
Followed the documentation of [installing shadcn in a Vue + Vite application](https://www.shadcn-vue.com/docs/installation/vite.html)

**We will need to use the CLI in order to use this library**, this was a thing that I have pointed as as benefic to the Shadcn approach, the ability to just copy and paste the components, but now I found that the vue library only allow to install via CLI

VUE Shadcn             |  React Shadcn
:-------------------------:|:-------------------------:
VUE Shadcn |  React Shadcn

As you can see, the React library has a guide to install the components manually, and in most cases is the same steps
- Download the Radix primitive component
- Create a file with some code that use this component on the Shadcn way

I don't see this as a problem, but we will need to have a new `components.json` file at the root of the project that is needed to use the CLI. 

**About theming**, there has a [documentation](https://www.shadcn-vue.com/docs/theming.html#adding-new-colors) to do it, we can iterate more on this.

### How to test if it works?
- Go to any component and put `bg-destructive` and see if it changes the BG.
- Install any component with the CLI and try to render it. ( installed components will be placed on the `/components/ui` folder

### This change can impact another functionality?
It can change the current style, but I think this is not a concern at the moment.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -1,26 +1,80 @@
+const animate = require("tailwindcss-animate")
+
 /** @type {import('tailwindcss').Config} */
-export default {
+module.exports = {
+  darkMode: ["class"],
+  
   content: [
-    "./index.html",
-    "./src/**/*.{vue,js,ts,jsx,tsx}",
-  ],
+    './pages/**/*.{ts,tsx,vue}',
+    './components/**/*.{ts,tsx,vue}',
+    './app/**/*.{ts,tsx,vue}',
+    './src/**/*.{ts,tsx,vue}',
+	],
+  
   theme: {
-    extend: {
+    container: {
+      center: true,
+      padding: "2rem",
       screens: {
-        'xs': '480px',
+        "2xl": "1400px",
       },
+    },
+    extend: {
       colors: {
-        'primary': '#7689F1',
-        'secondary': '#FFD46A',
-        'dark-gray': '#4B4D5B',
-        'light-gray': {
-          DEFAULT: '#E6E9F8',
-          250: '#F8F8FD',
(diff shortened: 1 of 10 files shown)


WaldemarLehner:
Please merge `rewrite` into this branch to fix the Pipeline.

tailwind.config.js
```diff
@@ -1,26 +1,80 @@
...
+    './src/**/*.{ts,tsx,vue}',
+	],
+  
   theme: {
-    extend: {
+    container: {
+      center: true,
+      padding: "2rem",
       screens: {
-        'xs': '480px',
```
JulianDietzel, line 10: Why was this removed?

JulianDietzel:
Apart from the question why the "xs" breakpoint was removed, this looks good to me :+1: 

We should then address adjusting the theming according to #309 in a follow up PR.

tailwind.config.js
Dosbodoke: The Shadcn CLI have changed the `tailwind.config.js` automatically. But do you think it's really necessary to add one more breakpoint for smaller devices? Shouldn't we follow [the default](https://tailwindcss.com/docs/screens#configuring-custom-screens) of tailwind?

JulianDietzel: @bastislack added it in #308 and I suppose he needs it. But then also he can add it back when / if he needs it. So I suppose it is fine for now.

PR #315, opened 2023-12-04
feat(ci): Create script to fetch Tricks and Combos from a Google Sheet  to store them in YAML Files

WaldemarLehner:
Okay… after way too many commits to try to fix the pipeline… here it is :D 

There might still be some weird edge-cases, but I believe that this should work as is.

Pipeline has a manual trigger, could add cron jobs on a daily basis or whatnot later on.

There are two ts scripts that form the "heart" of the PR:

`fetchFromSheet.ts` - Calls Google Sheet's API to retrieve TSV Tables. Deletes old YAML files and replaces with the new set inferred from the Sheet.

`createPrForSheetsIfNeeded.ts` - Checks the Git Diff and creates, updates or deleted a PR as needed.

Oh… and one more thing:
Here's the Link to the sheet :) 
https://docs.google.com/spreadsheets/d/1h1PGIrOPx16KH8IOiMh3SD1bSh1xIWFFrSL84Xyxj38/edit


.github/workflows/syncsheet.yml
```diff
@@ -0,0 +1,72 @@
...
+
+            # Create it if it is missing
+            git checkout -b chore/sheets
+            git branch --set-upstream-to=origin/chore/sheets chore/sheets || true
+          fi
+          git checkout chore/sheets
+
+          # This can fail if the branch is not on remote. 
+          # If that's the case, we don't care. It will be created on commit / push
+          git pull || true
```
JulianDietzel, line 53: What does appending `|| true` do?

.github/workflows/syncsheet.yml
```diff
@@ -0,0 +1,72 @@
...
+
+          # This can fail if the branch is not on remote. 
+          # If that's the case, we don't care. It will be created on commit / push
+          git pull || true 
+
+          # If here: Branch exists. But might still have a bad state.
+          # Merge from rewrite and favour their (rewrite) changes on conflicts
+          git rebase origin/${TARGET_HEAD_REF} -X theirs
+          
+          git status
```
JulianDietzel, line 59: This is just here to provide the developer with information right?

.github/workflows/syncsheet.yml
```diff
@@ -0,0 +1,72 @@
...
+
+          # If here: Branch set up and will (hopefully :P) not create any issues on changes
+      - name: Install Dependencies from Lock-File
+        run: npm ci
+      - name: Sync with the Sheet
+        run: npx vite-node ./scripts/fetchFromSheet.ts
+        # Sync with the Sheet has created some changes. For the next step these changes have to be staged,
+        # otherwise the diffing will not work
```
JulianDietzel: If this comment is true then why is `git add src/data` run in the next step. Also, on checking very briefly (I might have just missed it) the content of the `fetchFromSheet.ts` script, it doesn't seem like the script stages any changes. Is this comment incorrect, or unclear or am I mistaken?

.github/workflows/syncsheet.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Add YAML Changes to Git
+        run: |
+          git add src/data
+          npm run lint:staged -- --allow-empty
```
JulianDietzel, line 72: is the  `--` with nothing following it intentional?

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+    { sheetId: '1h1PGIrOPx16KH8IOiMh3SD1bSh1xIWFFrSL84Xyxj38', sheetGid: '0', name: 'Tricks' },
+    {
+      sheetId: '1h1PGIrOPx16KH8IOiMh3SD1bSh1xIWFFrSL84Xyxj38',
+      sheetGid: '2129196972',
+      name: 'Combos',
+    },
+    {
+      sheetId: '1h1PGIrOPx16KH8IOiMh3SD1bSh1xIWFFrSL84Xyxj38',
+      sheetGid: '738540593',
+      name: 'Videos',
```
JulianDietzel: Move Ids and Gids to out of the script and into environment variables. If the "name" values have to match values in the Google sheet move to the environment variables as well. If not and they are just needed for the script itself leave them as is. (It very strongly looks to me like the names are not used for the request and should not be converted to env-variables)

src/data/tricks/example2.yaml
JulianDietzel, line 1: 1. This trick has no comment at the top that it was generated automatically and has "#keep" at the end. Why and what does that mean?
2. If not answered by the first question: Given the trick was autogenerated, how did it change the "id". Shouldn't that be the defining attribute, or, to use the language of databases, the primary key?

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+  const response = await fetch(
+    new URL(
+      `https://docs.google.com/spreadsheets/d/${sheetId}/export?format=tsv&id=${sheetId}&gid=${sheetGid}`
+    )
+  );
+  if (!response.ok) {
+    throw new Error(`Response was not ok. SheetId: ${sheetId}, SheetGid: ${sheetGid}`);
+  }
+  const asText = await response.text();
+  const lines = asText.split('\r\n');
```
JulianDietzel, line 58: Does Google Sheets really still do the `\r\n`? I suppose you had your reasons, but it is surprising

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+  const cells = lines.map((e) => e.split('\t'));
+  return cells;
+}
+
+console.log(chalk.blue('Fetching CSVs from Google Sheets...'));
+
+const [tricksCells, combosCells, videoCells] = await Promise.allSettled(
+  [
+    { sheetId: '1h1PGIrOPx16KH8IOiMh3SD1bSh1xIWFFrSL84Xyxj38', sheetGid: '0', name: 'Tricks' },
```
JulianDietzel: Break into multiple lines for visual continuity with other list items.

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+      sheetGid: '2129196972',
+      name: 'Combos',
+    },
+    {
+      sheetId: '1h1PGIrOPx16KH8IOiMh3SD1bSh1xIWFFrSL84Xyxj38',
+      sheetGid: '738540593',
+      name: 'Videos',
+    },
+  ]
+    .map(async (e) => [e.name, await fetchTsvFromGoogleSheets(e.sheetId, e.sheetGid)] as const)
```
JulianDietzel: Rename "e" to "table" perhaps?

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+    .map((e) =>
+      e
+        .then((e) => {
+          console.log(chalk.green(`Fetched ${e[0]}, found ${e[1].length - 1} rows.`));
+          return e[1];
+        })
+        .catch((e) => {
+          console.log(chalk.red(`Failed to fetch a table; ${e}`));
+          throw e;
```
JulianDietzel: eeeeee. Please refactor to variable names that indicate the content of the variables

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+
+if (tricksCells.status === 'rejected') {
+  console.log(chalk.red('Failed to get Tricks from Sheets'));
+  console.log(chalk.gray(tricksCells.reason));
+  exit(1);
+}
+if (combosCells.status === 'rejected') {
+  console.log(chalk.red('Failed to get Combos from Sheets'));
+  console.log(chalk.gray(combosCells.reason));
+
```
JulianDietzel, line 107: Remove empty line

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+  }
+  const colLen = rowsWithHeader[0].length;
+  if (colLen === 0) {
+    console.log(chalk.red('Failed to get Data from Sheets'));
+    console.log(chalk.gray('A table does not contain any columns'));
+    exit(1);
+  }
+
+  // Missing cells are bad, additional cells are fine and can be ignored
+  if (rowsWithHeader.some((e) => e.length < colLen)) {
```
JulianDietzel, line 132: 1. Fine for now I suppose. If you feel like it, it could be really nice to tell the developer on which specific lines information is missing.
2. How do rows end up with too little columns? Is it just that the last column contains no data? Because then this could lead to valid combos or videos being rejected because there either are not tips for a combo or there is not end time provided for a video. Correct me if I am wrong here. Tricks should be fine by this logic as your last column for the tricks always contain a value in your sheet.

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+  return (
+    rowsWithHeader
+      // Skip the header row
+      .filter((_, index) => index > 0)
+      .map(
+        // create [headerName, cellValue] Tuple array
+        (row) =>
+          header
+            .map((head, i) => [head, row[i]] as const)
+            .filter(([_, val]) => val.trim().length > 0)
```
JulianDietzel, line 149: Am I correct that this filters out empty cells?

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+      // and turn the tuple array into {headerName: cellValue} entries
+      .map((e) => Object.fromEntries(e))
+  );
+}
+
+const [trickRawObjects, combosRawObjects, videoRawObjects] = [
+  tricksCells,
+  combosCells,
+  videoCells,
+].map((e) => turnRowsIntoObjects(e.value));
```
JulianDietzel, line 160: Nitpick or more of a question from someone less experienced with JS:
Is this the preferred or more optimal way to write this. The simple alternative of writing
```js
const trickRawObjects = turnRowsIntoObjects(tricksCells);
```
three times would use up two fewer lines of code and reduce the risk of messing up the order of either the inputs or the outputs.

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+
+// Videos need to be done first so they can be used in Trick and Combo-Objects
+// We create a Lookup-Map instead of an array here.
+
+const videoLookup: Record<string, { link: string; startTime: number; endTime: number }[]> = {};
+
+videoRawObjects.forEach((e) => {
+  const key = `${e.category}-${e.id}`;
```
JulianDietzel, line 168: What do you think of changing "category" to "stickableCategory"? Not sure myself but could be better for internal consistency

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+  combosCells,
+  videoCells,
+].map((e) => turnRowsIntoObjects(e.value));
+
+// Videos need to be done first so they can be used in Trick and Combo-Objects
+// We create a Lookup-Map instead of an array here.
+
+const videoLookup: Record<string, { link: string; startTime: number; endTime: number }[]> = {};
+
+videoRawObjects.forEach((e) => {
```
JulianDietzel, line 167: Refactor "e" to "video"

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+    variationOf: e.variationOf ? e.variationOf.split(',').map((e) => Number(e.trim())) : [],
+    tips: (e.tips ?? '')
+      .split(';')
+      .map((e) => e.trim())
+      .filter(Boolean),
+    videos: videoLookup['trick-' + e.id],
+    difficultyLevel: Number(e.difficultyLevel),
+    showInSearchQueries: e.showInSearchQueries === 'TRUE',
+    dateAddedEpoch: new Date(e.dateAddedIso8601).getTime(),
+    yearEstablished: e.yearEstablished ? Number(e.yearEstablished) : undefined,
```
JulianDietzel: Refactor "e" to "trick"

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+    id: Number(e.id),
+    recommendedPrerequisites: e.recommendedPrerequisites
+      ? e.recommendedPrerequisites.split(',').map((e) => Number(e.trim()))
+      : [],
+    variationOf: e.variationOf ? e.variationOf.split(',').map((e) => Number(e.trim())) : [],
+    tips: (e.tips ?? '')
+      .split(';')
+      .map((e) => e.trim())
+      .filter(Boolean),
+    videos: videoLookup['trick-' + e.id],
```
JulianDietzel: For "recommendedPrerequisites" and "variationOf" it is made sure that they always have a list as their value, even if that list is empty. Here, for "videos", if there is no lookup element, "videos" will not contain an empty list but be undefined. Is this intentional. If not, then I would advocate for making the system consistent. Either by changing the video default to a list or by changing the default of the other two to "undefined"

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+      .map((e) => e.trim())
+      .filter(Boolean),
+    dateAddedEpoch: new Date(e.dateAddedIso8601).getTime(),
+    yearEstablished: e.yearEstablished ? Number(e.yearEstablished) : undefined,
+    tricks: (e.tricks ?? '')
+      .split(',')
+      .map((e) => e.trim())
+      .filter(Boolean)
+      .map(Number),
+    videos: videoLookup['combo-' + e.id],
```
JulianDietzel, line 240: As in the comment before: Is there a reason for this being "undefined" instead of an empty list when there are no values? If not, probably change this

scripts/fetchFromSheet.ts
```diff
@@ -0,0 +1,333 @@
...
+      .join('\n') +
+    '\n\n' +
+    yamlContent
+  );
+}
+
+console.log(chalk.blue('Writing new Trick files...'));
+await Promise.all(
+  trickObjects
+    .map((e) => (e.success === true ? e.data : (undefined as never)))
```
JulianDietzel, line 346: What does this line do excatly / why is it needed?

scripts/createPrForSheetsIfNeeded.ts
```diff
@@ -0,0 +1,165 @@
...
+  return output
+    .split('\n')
+    .filter((e) => e.trim().length > 0)
+    .map((e) => {
+      const [operation, file] = e.split('\t');
+      return { operation: getOperations(operation[0]), file };
+    });
+}
+
+// Tries to find a Pull Requests with "automatic", "sheet-update" Tags that belongs has its HEAD on the current branch.
```
JulianDietzel: "[...] that belongs has its HEAD [...]" is not a proper sentence.

JulianDietzel (changes requested):
Thanks for all the work. I added a bunch of comments, some of them are just questions for clarification or asking about stuff I don't understand.

I am no git or GitHub-Actions expert (by any means) so I can't guarantee that there aren't any mistakes in the code but the general way you are going about the thing seems good and reasonable, checking already for a bunch of edge cases. :heart:

src/data/tricks/example2.yaml
JulianDietzel: Never mind, I get the "#keep" now after reading the scripts. But then, shouldn't we consider removing these example tricks now or at least remove the "#keep" so that they are removed the first time we properly run the action?

.github/workflows/syncsheet.yml
WaldemarLehner: Ye. Its just for debugging purposes

.github/workflows/syncsheet.yml
WaldemarLehner:
```sh
commandThatWillFail
echo $?
# something not 0
commandThatWillFail || true
echo $?
# 0
```

In short, if a command fails, the or operator will cause the second command to be invoked.

true is a command that always returns status 0

.github/workflows/syncsheet.yml
WaldemarLehner: Yes. This is as expected. 

It passes everything after the double dash to the script that is defined in the scripts section.

src/data/tricks/example2.yaml
WaldemarLehner: Ye. I kept it here for now to demonstrate the keep feature

scripts/fetchFromSheet.ts
WaldemarLehner: Yup. The tsv export uses carriage return line feed endings.

I first split on newlines, but then had a carriage return at the end of lines

scripts/fetchFromSheet.ts
WaldemarLehner: Empty cells are still passed correctly and are understood as cells.

It's just a case of defensive programming, just for the case that we would get oddly formatted files from sheets

scripts/fetchFromSheet.ts
WaldemarLehner: Basically, yes.

The column header is used as the key, the cell as the value 

This removes all empty cells from the object that gets constructed.

scripts/createPrForSheetsIfNeeded.ts
WaldemarLehner: Tldr: don't code at 3am :D

.github/workflows/syncsheet.yml
WaldemarLehner: image

scripts/fetchFromSheet.ts
WaldemarLehner: The names are not used, but we need to know which Request has the data from the Videos-Sheet, which one has the Combos-Sheet, etc.
I ill move the IDs out of there

.github/workflows/syncsheet.yml
JulianDietzel: Interesting and good to know. Thanks for the explanation

JulianDietzel:
When merging we should remember to updated the appropriate section in the Wiki (it still has a note, that this script is not written and it is undefined when / how the script is triggered):
https://github.com/bastislack/highline-freestyle/wiki/Backend-Overview#1-from-the-google-sheet-to-the-app

scripts/fetchFromSheet.ts
WaldemarLehner: Im confused about which statement would be `undefined` instead of `[]` here?


`tricks`? This is a `[]` if no values are present.

scripts/fetchFromSheet.ts
WaldemarLehner: There is a small bug as it's missing `.filter(e => e.success)`

The response here is a composed type of 
```ts
type X = {
  success: true,
  data: ...
} | {
  success: false,
  error: ...
}
``` 
TS only lets me access `data` if it understands that `success` is `true`.

Unfortunately, TS does not understand that a `filter( e => e.success)` results in that condition.
image

And that's where the *cursed* map comes into play. Casting a field to `never` basically tells TS that something  will *never* happen.

You usually see it on functions that will loop for ever, or can only throw something.
e.g.
image
image

so we map the data is success, or undefined (or anything, really) if its not successful, but because we filtered for successful data, that undefined never happens — but we need to tell TS that it never will with the `never` cast

scripts/fetchFromSheet.ts
WaldemarLehner: renamed to `csvTrick`

scripts/fetchFromSheet.ts
WaldemarLehner: They can have optional values, that's fine :) 

Keep in mind we do not convert to the "final" Tricks / Combos here, but to the YAML Representation.

The YAML Schemas are defined as follows:

```ts
// src/data/viteSchemaGenerator.ts
const ShortIdArray = z.array(z.number().int().min(0));

export const YamlTrickTableSchemaZod = DbTricksTableZod.omit({
  trickStatus: true,
  recommendedPrerequisites: true,
  variationOf: true,
}).and(
  z.object({
    recommendedPrerequisites: ShortIdArray.optional(),
    variationOf: ShortIdArray.optional(),
  })
);

export const YamlComboTableSchemaZod = DbCombosTableZod.omit({
  comboStatus: true,
  tricks: true,
}).and(
  z.object({
    tricks: ShortIdArray.nonempty(),
  })
);
```

scripts/fetchFromSheet.ts
WaldemarLehner: nvm. saw comment above. Ye, thats not an issue as the (YAML) Schema allows optional ones here.

.github/workflows/syncsheet.yml
WaldemarLehner: The diff check is done using `git diff --name-status HEAD`, which compares the staged files with the last commit.

Iirc I had to use this approach (over diffing the unstaged files) because `lint-staged` (as the name implies :P) would otherwise not lint and format the newly generated files.

Here's the relevant segment:

```ts
// scripts/createPrForSheetsIfNeeded.ts

console.log(chalk.blue('Looking for changes since last detection:'));

// This checks the currently staged files.
const localDiff = parseDiffOutput((await $`git diff --name-status HEAD`).stdout).filter(
  // We only care about data-Changes. There *shouldnt* be any other changes, but hey.. you never know :P
  (e) => e.file.startsWith('src/data')
);
console.log(localDiff);
```

scripts/createPrForSheetsIfNeeded.ts
WaldemarLehner: fixed

JulianDietzel: Still not the most proper English sentence but I suppose it is super good enough

.github/workflows/syncsheet.yml
JulianDietzel: I think we are not on the same page here. I am referring to this part of the comment: "For the next step these changes have to be staged, otherwise the diffing will not work".

When reading this I assumed that "next step" refers to the next action (or whatever the proper name is) in this file ("Add YAML Changes to Git", line 64). And with this reading, the comment is incorrect, since in `./scripts/fetchFromSheet.ts` no files are being staged if I am not mistaken. The staging seems to happen _after_ the script (`./scripts/fetchFromSheet.ts`) has terminated in the next action (line 64 of this script we are commenting on).

If I am correct, that the staging happens later in this file and not in `./scripts/fetchFromSheet.ts`, please changed the wording so that it is more clear what happens when. If I am mistaken on this, please correct me.

Also, I just noticed "Sync with the Sheet has created some changes" is not a proper English sentence and needs to be corrected as it is not exactly clear what you are trying to say.

scripts/fetchFromSheet.ts
JulianDietzel: That makes sense. It is not perfectly consistent but not an issue either.

scripts/fetchFromSheet.ts
JulianDietzel: I think I mostly understand it and this seems alright though slightly cursed. Since this is really specific please add a comment for clarity (even if it is just something as simple as "This .map acts like .filter(e => e.success) but was used for compatibility with Typescript")

WaldemarLehner: done

.github/workflows/syncsheet.yml
WaldemarLehner: Changed to 
```yaml
        run: npm ci
      - name: Fetch Changes from Google Sheet
        run: npx vite-node ./scripts/fetchFromSheet.ts
        # Fetching from the Google Sheets might have created some changes in the YAML Files.
        # Stage any YAML Files. The --allow-empty is needed here, otherwise husky will try to 
        # create an empty commit, fail, and as result, break the pipeline.
      - name: Add YAML Changes to Git
        run: |
          git add src/data
          npm run lint:staged -- --allow-empty
          git status
        # Staged changes are used in the createPrForSheetsIfNeeded.ts Script to determine if a new PR and/or commit is needed.
      - name: Run the PR Procedure
```

JulianDietzel: That is more understandable :+1: . Also, I just realized that the misunderstanding was mostly on me (and a little silly). I assumed that the comment was talking about the line above it not the action bellow. I'm not quite sure how I got that notion :D

WaldemarLehner: >  I assumed that the comment was talking about the line above it not the action bellow. I'm not quite sure how I got that notion :D

I blame python multiline comments :D

WaldemarLehner:
> When merging we should remember to updated the appropriate section in the Wiki (it still has a note, that this script is not written and it is undefined when / how the script is triggered): https://github.com/bastislack/highline-freestyle/wiki/Backend-Overview-(WIP)#1-from-the-google-sheet-to-the-app — @JulianDietzel 

This links me to an empty page. Was that expected? The (WIP) would make me believe that there is already a page set up.

JulianDietzel:
> > When merging we should remember to updated the appropriate section in the Wiki (it still has a note, that this script is not written and it is undefined when / how the script is triggered): https://github.com/bastislack/highline-freestyle/wiki/Backend-Overview-(WIP)#1-from-the-google-sheet-to-the-app — @JulianDietzel
> 
> This links me to an empty page. Was that expected? The (WIP) would make me believe that there is already a page set up.

Yes, the page already exists. I changed the the name of the page recently though so the link went dead. I updated the link in the original comment.

src/data/tricks/example2.yaml
JulianDietzel: Demonstration purposes are nice but I think it would be better to add an explanation to the wiki than risking the example tricks slipping into the production build. I'd suggest deleting them.

scripts/fetchFromSheet.ts
JulianDietzel: Please refactor this one. The other ones I originally marked are fine, but the nesting and shadowing of the name "e" as a consequence here are lead to poor readability.

JulianDietzel (changes requested):
I marked all previous conversations that I now deem to be ok as resolved, please address the remaining two comments.

scripts/fetchFromSheet.ts
WaldemarLehner: Changed to 
```ts
.map(
  async (googleTableData) =>
    [
      googleTableData.name,
      await fetchTsvFromGoogleSheets(googleTableData.sheetId!, googleTableData.sheetGid!),
    ] as const
)
.map((nameAndDataPromise) =>
  nameAndDataPromise
    .then(([tableName, tableData]) => {
      console.log(chalk.green(`Fetched ${tableName}, found ${tableData.length - 1} rows.`));
      return tableData;
    })
    .catch((err) => {
      console.log(chalk.red(`Failed to fetch a table; ${err}`));
      throw err;
    })
)
```

src/data/tricks/example2.yaml
WaldemarLehner: Removed #keep-Annotated Entry

JulianDietzel (approved):

PR #318, opened 2023-12-06
feat(ui): Add proper navbar

JulianDietzel:
Adds a proper responsive navbar for both desktop (side) and mobile (bottom). Replaces PR #303.

The navbar routes to
- `/tricks`
- `/combos`
- `/glossary`
- `/about`

The navbar has a language selector to select between English, Spanish and French.

Here is how the navbar looks. Dropdown in the mobile view can be opened with the three dots on the right and is not always visible:
| Desktop | Mobile |
| --- | --- |
| Bildschirmfoto am 2023-12-06 um 20 37 26 | Bildschirmfoto am 2023-12-06 um 20 36 56 |

I had to install `lucide-vue-next` despite us already having a library for icons. This is due to an issue with shadcn's dropdowns. If the library is not installed and a dropdown is used, a runtime error occurs, pointing to to the missing package.


WaldemarLehner (approved):
Ye. Looks good to me :) 

I made a minor change to the bottom nav, just to give it a bit smoother experience :) 

Personally, I wouldnt split up the Dropdown in so many components, but the way its set up should work regardless.

JulianDietzel:
Found a bug: When the dropdown is open in the mobile navbar and the screen gets resized so that the side navbar becomes active, the dropdown stays open and visible instead of disappearing with the bottom navbar. Hence the change to a draft. I'll fix this first, then convert it back to a proper PR.

WaldemarLehner:
I don't see this as an issue tbh.

It's an edge case that happens on a resize that has no negative side effect.

If the drop down stays open, even after clicking somewhere else in the app,
then yeah. That has to be resolved.

But if its just open in this very niche edge case? I'd classify that as a
won't fix

JulianDietzel:
Agreed, won't fix for now

Issue #161, opened 2022-01-14
New Logo


bastislack:
I talked to Gaelle and she is motivated to design a new logo for the app as soon as we have chosen our main theme colors for the app.

hell3nacht:
hey and hello, @JulianDietzel and i played around with some logo ideas (picture below, not everything is to be taken seriously). due to the variety: feedback or preferences are highly appreciated! luzie :) 

<img width="968" alt="2024_01_06_logovarianten" src="https://github.com/bastislack/highline-freestyle/assets/132842524/9f8d95fb-7dfd-4a09-969e-450a3d38fec9">

JulianDietzel:
To add my two cents, partially already discussed with @hell3nacht, here are the ones I like best out of the ones listed:

| Number | How much I like it (o, +, ++) | Comment |
|--------|--------|--------|
| 1 | + |  Originally I found the lines to be to thin. But they might fit the look of the app after all |
| 6 | + | Basic but works |
| 8 & 9 | ++ | Both work quite well. I have a slight preference for the thinner 8. Perhapse a less round font could look good aswell. |
| 11 & 12 | ++ | I like how "Highline" and "Freestyle" are of the same width. 12 looks a little better to me. The only think I dislike a little is how "Freestyle" is smaller than "Highline" but that's not big enough an issue to discard the idea. |
| 13 | ++ | Like it aswell, slightly better than 6 |
| 25 | o | It's ok but I like the others better | 
| 29 | ++ | I still really like the look of this one but do recognize that "Highline" is just barely legible | 
|  |  |  | 
| 21 & 22 |  | I like the usage of the initials for the small version. The only thing that bothers me is how the only slanted line is sloping downwards instead of up and down / upwards |

We were tinkering with completly different ideas where we had line drawings of an animal we used for the small and large logo (like for example a sloth, a kiwi or, after some getting used to it my current favourite, a rooster). It had a really strong energy to it. I personally would love seeing this idea explored further but myself don't have any time to spend on this.

If you don't get any feedback from the others within the next few days feel free to just go with whatever you think would be cool :+1:.

If you want to work with colors there is a wiki entry on the colors we are currently using [here](https://github.com/bastislack/highline-freestyle/wiki/Colors). The colors so far are very close to what they used to be before the rewrite so the _might_ change in the foreseeable future. I opened issue #322 to discuss what colors we want to use in the future (feel free to join / start that discussion as well).

hell3nacht:
[...]

PR #331, opened 2024-01-26
feat: add adaptive favicon

JulianDietzel:
Last addition to wrap up issue #161.

- Updated the favicons so that the logo is larger relative to the image size (effectively make the logo larger in the browser tab)
- Updated the favicon names
- Installed `vueUse` to both get the browser theme (dark / light) and the functionality to set the favicon
- Made the favicon adaptive to the browser theme (turning dark when the theme is light and vise versa)


WaldemarLehner (changes requested):
Assuming the icon is a vector graphic, this can be done without any JS by using a SVG for the favicon which contains a Style-Tag which queries the `prefers-color-scheme: dark` Property.

See https://www.youtube.com/watch?v=tXGG9UCM5lE for a walkthrough on how to achieve this

JulianDietzel:
Since all changes of this PR need to be / should be reverted for the suggested solution I will deleted this PR and will open an new branch and PR (#333)

PR #333, opened 2024-01-27
feat: add adaptive favicon

JulianDietzel:
Modified the favicon so that it is a) properly displayed and b) adapts to the theme of the browser (the icon is dark when the theme is light and vice versa). This should warp up issue #161 and is the second attempt at this, with #331 being the first but rejected one. 

The more elegant method used to make the favicon adapt to the browser theme in this PR was recommended by @WaldemarLehner in #331 . It relies on styling in the svg-file itself.


WaldemarLehner:
From purely looking at the changes this looks good 

Cannot test locally as my root partition just s.... the bed :(

JulianDietzel:
Sweet. It runs on both my Ubuntu Laptop (Firefox) and my Android Phone (also Firefox) so I thinks this should be [super good enough](https://fhm.nl/wp-content/uploads/2017/10/Schermafbeelding-2017-10-16-om-16.23.56-1024x554.png)

Issue #326, opened 2024-01-19
Introduce cookie settings


JulianDietzel:
> [!NOTE]
> Important follow up issue to #279. The following was moved to this separate issue to a) contain the scope of #279 and b) since this was not originally part of the issue. This should be addressed only once #279 is implemented.

Since in #279 video embeds are introduced we need a way to allow the user to consent to or decline the use of third party cookies. To do this, the following needs to be put in place:
- [x] A page (/part of a page) where the cookie settings can be set globally for the entire app. Adding a settings page is probably the most appropriate solution for this. This page should contain:
   - [x] A short description what these settings do (they allow us to show the videos but can give the embed provider the possibility to track the user. We ourselves don't track the user and don't place any cookies)
   - [x] A toggle for YouTube
   - [x] A toggle for Instagram
- [x] A system for storing the users preferences (some kind of in-browser storage)
- [x] A component which is shown instead of an embed if the user has not (yet) consented to cookies from the provider of the embed. This component should include:
   - [x]  A short description what this setting does (it allows us to show the video but can give the embed provider the possibility to track the user)
   - [x] The option to allow embeds from this provider permanently (/globally)
- [x] A short wiki entry for how this system works (where are the preferences stored)

WaldemarLehner:
I have made something similar for my website. We could reuse that logic if you want :) 

https://waldemarlehner.de/en/privacy/

JulianDietzel:
Started working on this. Current progress can be found on [feat/cookie-settings](https://github.com/bastislack/highline-freestyle/tree/feat/settings-page).

JulianDietzel:
Only some translations of the `video.en.json` file into Spanish and French, and a short wiki entry are still missing. Then this will be done.

JulianDietzel:
Implemented in #338

Issue #279, opened 2023-09-17
Trick List - Trick Details


bastislack:
As a Freestyler, I want to see the information about a trick so that I understand how it works and how I can learn it.

Should...
* Displays in the following order
  * Name (Alias)
  * Technical Name
  * Level
  * Start and End position
  * Video(s)
  * description
  * tips
  * prerequisites
  * established by / date
  * variation of
* Allow navigation to prereq. tricks and base trick (variation of)
* Allow navigation to trick overview
* All videos can be accessed through a simple menu
* When info is missing, show that it is missing except for
  * variation of
  * prerequisites
  * established by / date

JulianDietzel:
I'll give this a shot and work on the page.

JulianDietzel:
All progress on this is pushed onto [feat/rewrite-trick-details](https://github.com/bastislack/highline-freestyle/tree/feat/rewrite-trick-details).

JulianDietzel:
Since embedding the videos is a bit of a pain and taking a while here is a brief look at the current state:

<img src="https://github.com/bastislack/highline-freestyle/assets/33333283/078f5100-3155-47e3-ad78-e70b73bb0a39" width="200" />

<img src="https://github.com/bastislack/highline-freestyle/assets/33333283/de6aa8c5-612a-4df3-ac8a-b50cd28ebad6" width="200" />

Feel free to give feedback if you already have any. What still needs to happen is
- Adding the proper video embeds (YouTube and Instagram. Vimeo if causes absolutely no additonal work)
- Integrating the trick difficulty somewhere into the design
- Integrating the info if the trick is an official one or not into the design
- Integrating the info if the trick is new into the design
- Internationalizing the page

What can happen is:
- Adding a list of all the variations of the trick (if there exist  any)
- Somehow adding the stick frequency to the recommended prerequisites

The personal / metadata section (stickFrequency, notes, isFavourite) will be added later, on the bottom of the page in issues #289, #293 and #294.

bastislack:
Nice, I really like the design with the icons for all the titles!
Are you struggling with both youtube and ig embeds? I remember that youtube was more or less smooth but ig was a rough process in the first version of the app.

bastislack:
The info whether a trick is an official one or not could be shown by a small badge next to the main title (trick name), or after the technical name of the trick.

JulianDietzel:
I'm glad you like the design, thanks :).

[...]

PR #334, opened 2024-02-03
feat: trick details page

JulianDietzel:
After a lot more work than expected, here is the PR addressing #279. All requirements of that issue were addressed except for the very last one: "When info is missing, show that it is missing except for established by / date". This field is simply missing if there is no info. I found this to be a much cleaner look and nothing being lost from not displaying that the info was missing - it now simply _is_ missing. The page looks good both on mobile and on desktop.

#### Errors
Additionally there are error pages for all errors that I could conceive of that could occur when accessing the page. All error pages have fitting and proper error codes, titles and descriptions. All these texts have been translated into Spanish and French as well using DeepL and the little French knowledge that I have. I also created a custom vector graphic for the error page - me, taking a leashfall - which I thought was cute.

#### Translation
Lastly all text of the page and the video components (exept for the content) is translated into Spanish and French, again using a mixture of DeepL, Linguee, dict.cc and what is left of 5 years of French in highschool.

#### Sorryyy
I know this PR is excessively large, sorry for that. With the video components I got quite large in scope. I'll try to keep my next PR smaller.

#### Comments
I am open and happy about feedback both in regards to the ui and the code. If you have major changes that you want to see done to the UI I would suggest opening a new issue though, so we can make the changes  more incrementally and not bloat this PR any further. If I do any weird stuff code wise, please let me know though. Especially having the TrickDetailsContent component separated from TrickDetails feels slightly unorthorthodox though it made the code look quite a lot cleaner since the checking of the URL parameters and the loading of the trick are separated into different files this way.

## Remaining issues
- There is a bug with the navbar that makes it not highlight the "Trick" tab as active even when the Trick Details page is open. An new issue will need to be created for this. I'll try to do this soon.
- Since the videos are embeds it is both our legal and an ethical obligation to ask the user whether they want to allow the embedding of third party code (Instagram and YouTube). I already created an issue for this: #326.
- Whether the trick name will be displayed in the Header or not is currently still in discussion in #279.

## How it looks:
| Standard (top) | Standard (bottom) | Error |
| --- | --- | --- |
| Bildschirmfoto am 2024-02-02 um 21 04 46 | Bildschirmfoto am 2024-02-03 um 17 34 44 | Bildschirmfoto am 2024-02-03 um 17 36 22 |

| Desktop (top) | Desktop (bottom) |
| --- | --- |
| Bildschirmfoto am 2024-02-03 um 17 35 41 | Bildschirmfoto am 2024-02-03 um 17 35 46 |


src/assets/img/leashfall.svg
WaldemarLehner, line 1: Looks f**ing sick on the 4XX Pages! :heart:

src/components/ErrorInfo.vue
```diff
@@ -0,0 +1,92 @@
...
+  return window.history.length > 2;
+}
+</script>
+
+<template>
+  <div class="w-full flex flex-col items-center gap-4">
+    <div class="w-full flex flex-col items-center md:flex-row md:gap-3">
+      <img
+        v-if="!hideImage"
+        src="@/assets/img/leashfall.svg"
```
WaldemarLehner: I am not entirely sure if this would still work when built. 
It works in dev-mode because vite proxies the File System, but would fail in build mode.

Either `import LeashFallUrl from "@/assets/img/leashfall.svg?url` and replace with `:src=LeashFallUrl` or put the Svg into the public folder. Personally I would really gravitate towards the first solution.

src/components/ErrorInfo.vue
```diff
@@ -0,0 +1,92 @@
...
+import messages from '@/i18n/error';
+import Button from '@/components/ui/button/Button.vue';
+import { Separator } from '@/components/ui/separator';
+import { Collapsible, CollapsibleContent, CollapsibleTrigger } from '@/components/ui/collapsible';
+
+defineProps<{
+  title: string;
+  code?: number;
+  description?: string;
+  debug?: string;
```
WaldemarLehner, line 16: How would you feel about debug being a slot instead?

src/components/stickable/trick/TrickDetailsContent.vue
```diff
@@ -0,0 +1,286 @@
...
+      </Section>
+
+      <!-- Videos -->
+      <Section
+        v-if="trick.videos !== undefined && trick.videos.length >= 1"
+        class="bg-secondary w-full"
+      >
+        <VideoCarousel :videos="trick.videos" />
+      </Section>
+      <Separator v-else />
```
WaldemarLehner, line 168: any reason this Separator is only here if no videos are present?

src/components/stickable/trick/TrickDetailsContent.vue
```diff
@@ -0,0 +1,286 @@
...
+              <li
+                v-for="variation in variationOfFull"
+                :key="variation.primaryKey[1] + '-' + variation.primaryKey[0]"
+                class="pb-1"
+              >
+                <RouterLink
+                  :to="'/tricks/' + variation.primaryKey[1] + '/' + variation.primaryKey[0]"
+                  class="underline"
+                >
+                  {{ variation.alias ? variation.alias : variation.technicalName }}
```
WaldemarLehner:
```suggestion
                  {{ variation.alias ?? variation.technicalName }}
```

src/components/video/Duration.vue
```diff
@@ -0,0 +1,49 @@
...
+  end?: number;
+}>();
+
+const i18n = useI18n({
+  messages,
+  useScope: 'local',
+});
+const { t } = i18n;
+
+function minutesSecondsFromSeconds(seconds: number): string {
```
WaldemarLehner, line 23:
```suggestion
/**
 * Takes a number of seconds and pretty prints in mm:ss notation
 * e.g. 54 -> 00:54
 *      125 -> 02:05
 */
function minutesSecondsFromSeconds(seconds: number): string {
```

src/components/video/Duration.vue
```diff
@@ -0,0 +1,49 @@
...
+
+function numToString(number_: number, minLength: number): string {
+  let numAsString = number_.toString();
+
+  while (numAsString.length < minLength) {
+    numAsString = '0' + numAsString;
+  }
+
+  return numAsString;
+}
```
WaldemarLehner:
```suggestion
function numToString(number_: number, minLength: number): string {
  return String(number_).padStart(minLength, "0");
}
```

WaldemarLehner: also… `zeroPadNumber` or something like that is probably a better name

src/components/video/VideoCarousel.vue
```diff
@@ -0,0 +1,74 @@
...
+  isCurrentVideoLast.value = currentVideoIndex.value == props.videos.length - 1;
+});
+
+function skipToNext() {
+  currentVideoIndex.value = Math.min(currentVideoIndex.value + 1, props.videos.length - 1);
+}
+
+function skipToPrev() {
+  currentVideoIndex.value = Math.max(currentVideoIndex.value - 1, 0);
+}
```
WaldemarLehner: nit: replace skip with go.
In this instance, we are not jumping over any videos, just going to the next / prev one.

src/components/video/platforms/YoutubePlayer.vue
```diff
@@ -0,0 +1,124 @@
...
+  const videoID: string = videoIdFromURL(url_);
+
+  let suffix: string = startTime_ || endTime_ ? '?version=3&loop=1&modestbranding=1&' : '';
+  suffix += startTime_ ? 'start=' + startTime_ : '';
+  suffix += startTime_ && endTime_ ? '&' : '';
+  suffix += endTime_ ? 'end=' + endTime_ : '';
+
+  console.log(prefix + videoID + suffix);
+
+  return prefix + videoID + suffix;
```
WaldemarLehner: use `URL()` instead

Also… why are the params suffixed with `_`?

src/components/video/platforms/YoutubePlayer.vue
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/**
+ * Parsing out the video ID from a YouTube url. The solution was take from
+ * https://stackoverflow.com/a/8260383
+ * @param url Youtube URL in any format (youtube.com, youtu.be, ...)
+ */
+function videoIdFromURL(url_: string): string {
```
WaldemarLehner: This looks incredibly scary :sweat_smile: 

Why is the param suffixed with `_`?

src/i18n/video/video.fr.json
```diff
@@ -0,0 +1,12 @@
+{
+  "duration": {
+    "to": "à",
+    "not-specified": "Durée non spécifiée"
+  },
+  "error": {
+    "only-youtube-and-insta": "Seules les vidéos YouTube et Instagram peuvent être affichées !",
+    "invalid-url": "URL vidéo non valide !",
+    "invalid-youtube": "L'URL YouTube fournie ne peut pas être analysée ! La vidéo ne peut pas être affichée."
```
WaldemarLehner, line 9:
```diff
- … !
+ …!
```

WaldemarLehner (changes requested):

WaldemarLehner:
I found that the current impl does not support youtube clips. I ill do a fix which lets the Youtube component also display Clips

src/i18n/video/video.fr.json
JulianDietzel: Thought the same until DeepL repeatedly put a space before exclamation marks. After a quick internet search it turns out this is actually a quirk of the French language and not a mistake.

src/components/video/platforms/YoutubePlayer.vue
JulianDietzel: Fixed in new commits of @WaldemarLehner.

src/assets/img/leashfall.svg
JulianDietzel: Thanks :heart: !

src/components/stickable/trick/TrickDetailsContent.vue
JulianDietzel: Yes. If there are videos they, with their light grey background, are visually separated from both the header and the following content as both of them have a white background. If there are no videos then the content placed immediately underneath the header. Since they are of the same background color a separator is now needed to,  separate them visually.

src/components/ErrorInfo.vue
JulianDietzel: Accepted your first suggestion and implemented it.

src/components/stickable/trick/TrickDetailsContent.vue
JulianDietzel: Accepted and also changed this in other places.

I do want to remark though, that there is a benefit to the "old" / my version. If for any reason the alias is empty ("") and not undefined, then my version would replace it with the technical name while with the nullish coalescing operator the empty string will be used. The alias should never be empty though and the nullish coalescing variant looks cleaner in the code.

src/components/video/platforms/YoutubePlayer.vue
JulianDietzel: Talked about it. Fixed by removing the suffix.

src/components/video/Duration.vue
JulianDietzel: Accepted and implemented both the code and the name change.

src/components/video/VideoCarousel.vue
JulianDietzel: Accepted and implemented

src/components/ErrorInfo.vue
JulianDietzel: I took a look at it. While not necessary I do see the appeal of making it a lot, allowing more complex content to be shown. The only issue is, that the debug info (and the collapsible with it) should only be shown if it is provided. But checking if slot content is provided seems possible though a little complicated.
I don't feel like putting in the time to put into this but if you feel like it, this seems to be the relevant GitHub thread to solve the issue:
https://github.com/vuejs/core/issues/4733

src/i18n/video/video.fr.json
WaldemarLehner: just confirmed with a french friend of mine :)

WaldemarLehner (approved):

PR #353, opened 2025-01-28
Fix form validations

### LINKED ISSUE #347: Can't select start and end positions when creating new user defined trick
# Issue
When trying to create a new trick via the `/tricks/new` route, changing the start and end positions to anything else than the default values, leaves the selection blank and shows the error "This is required" to the user.

# Expected behavior
When clicking a position in the dropdown dialogue, the position should be selected, shown in the field and no error should be shown to the user.

# Screenshot
Image

Dosbodoke:
Fixes #347

## What's the PR for?

Fix the select component that was not selection a trick at all.

## What was done to achieve it?

For the select itself it was a simple fix of just adjusting the property `@update:model-value`

During the refactor I noticed that the `zod` library was not being used in a ideal way and not integrated with the form library at all, so I refactored all that code, installed `@vee-validate/zod` and make them work seamlessly.


JulianDietzel:
Thanks for taking on this issue <3

Do you mean that the PR is still WIP on a functional level (will there be more changes to the code?) or that the description of it is still WIP?

Dosbodoke:
> Thanks for taking on this issue <3
> 
> Do you mean that the PR is still WIP on a functional level (will there be more changes to the code?) or that the description of it is still WIP?

I was referring primarly to the PR description, even tough I have one more key point that would like to solve on this issue, that is properly handling the zod error tranlations.

There is a well know library for this that is even cited on the zod official documentation

> [zod-i18n-map](https://github.com/aiji42/zod-i18n): Useful for translating Zod error messages.

Should I create a new issue or address it on this one?

src/i18n/error/error.en.json
```diff
@@ -3,7 +3,7 @@
   "home": "Home",
   "debug-info": "Debug info",
 
-  "INPUT_REQUIRED": "This is required",
+  "Required": "This is required",
```
WaldemarLehner, line 6: I dont have any strong opinions about REQUIRED vs Required, my only strong opinion is that it should be consistent :)

src/components/ui/customForm/TrickSelect.vue
```diff
@@ -36,7 +33,6 @@ const { t } = useI18n({
     :validate-on-change="true"
     :validate-on-input="true"
     :validate-on-blur="true"
-    :rules="validator as RuleExpression<unknown>"
```
WaldemarLehner, line 39: Why are the rules removed here?

src/components/ui/form/FormMessage.vue
```diff
@@ -1,9 +1,18 @@
 <script lang="ts" setup>
 import { ErrorMessage } from 'vee-validate';
 import { toValue } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+import messages_errors from '@/i18n/error.ts';
```
WaldemarLehner: nit: snake case?

JulianDietzel:
Talking to @WaldemarLehner we agreed that using zod-i18n-map would be a good idea. I personally would put it in a new PR. If you want to you can put it in this PR though. You probably know better what makes more sense anyway. So feel free to make the call.

WaldemarLehner (approved):
Haven't looked at the code in a loong while :D 

Didn't know you can pass in a specific message to zod for constraints. That's definitely an improvement over what we did w/ Preprocess before.

I have a few minor nits, which basically concern themselves w/ consistency.

src/i18n/error/error.en.json
Dosbodoke: I just didn't wanted to `z.string({ message: "INPUT_REQUIRED" })` as **Required** is the default error name evaluated by zod. We may create a new issue to fix all these error translation messages.

Dosbodoke: What do you think, should stick with **INPUT_REQUIRED** for now?

JulianDietzel: Since @WaldemarLehner didn't get back to me with his opinion and mine isn't all to strong on the matter I suggest we leave it as it is in the PR at the moment for now.

PR #363, opened 2025-02-17
Added brazilian portuguese translation

### LINKED ISSUE #352: Brazilian portuguese translation
**User story:**

As a brazilian freestyler, I want to be able to use the App even tough I don't know any foreing language.

Dosbodoke:
Resolves #352 

Besides the translations, I have made one important change.

Previously, the options to change the language were defined once for each navbar `NavbarBottom` for mobile and `NavbarSide` for desktop, now it's defined once on `src/util/locale.ts`.

### DIFF
--- a/src/components/navbar/NavbarBottom.vue
+++ b/src/components/navbar/NavbarBottom.vue
@@ -10,40 +10,18 @@ import {
 import { Icon } from '@iconify/vue';
 import { useI18n } from 'vue-i18n';
 
-import messages from '../../i18n/navbar';
+import messages from '@/i18n/navbar';
 import DropdownMenuGroup from '../ui/dropdown-menu/DropdownMenuGroup.vue';
-import { setNewLocale } from '@/util/locale';
+import { setNewLocale, LocaleInfos, type Locales } from '@/util/locale';
 
 const i18n = useI18n({
   messages,
   useScope: 'local',
 });
-
 const { t } = i18n;
-
-interface LocaleInfo {
-  locale: 'en' | 'fr' | 'es';
-  name: string;
-}
-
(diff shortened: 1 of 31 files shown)


Dosbodoke:
On my app, I have added a github test to check if all translation JSON has the same structure givin it the safety that the translations will not be broke on any language.

https://github.com/Dosbodoke/chooselife/blob/develop/.github/workflows/check-translations.yml

This my be a good enhancement here, tell me what you guys think about this.

JulianDietzel:
Thanks for the PR! I totally missed it (for some reason I didn't get an email for it). I'll take a look at it in the next couple of days.

JulianDietzel:
Thanks for all the work! Translating can be a pain. The changes you made to the Bottom- and Side-Navbar look good to me. And in terms of the translations you are the expert anyways :D.

I did like your suggestion of adding a GitHub test to check if all the translation files have the same structure initially. However, my understanding is that this will require all the translation files to have the exact same JSON structure, meaning that no entries can be missing. Please correct me if this assumption is wrong. If it is correct though this will mean that either no commits or no PR merges are possible where one or more translations are missing. This would be an issue if for example I add a text element to the App and would want to leave the Portuguese translation to you (in a later commit). This would no longer be possible.

It could be a consideration though for a GitHub-Action which is triggered before the App gets deployed.

I will accept this PR and therefore close it. Since this probably does not need to be discussed on GitHub, feel free to text me on Discord if you want to discuss the GitHub Action further.

Dosbodoke:
Yeah you're right, if the structure of the JSON files were different the job would fail. But it can be configured to not block on 'rewrite'/'develop' and block to 'main'

PR #367, opened 2025-03-06
Fixed validations for number input

### LINKED ISSUE #362: Can't create new trick: Incorrectly validated Difficulty field
# Description
When trying to create a new personal trick via the `/tricks/new` route, setting a difficulty results in a validation error. The message reads "Expected number, received string". As a result it is not possible to create a new personal trick at all unless no value is in the difficulty field.

I believe this issue is new since #353. So maybe @Dosbodoke could take a look.

# Expected behavior
There should only be an error it there is actually something else in the field than a number. If there is a number in the field the number should be used for the difficulty, there should be no error, and the user should be able to create the trick.

(While at it, it might be nice to only allow inputting numbers in the first place (ignoring everything else all together). But this is not necessary.)

# Screenshot

Image

Dosbodoke:
Fixes #362 

# How to test

Go to the new trick page and try to mess with difficulty and year fields.

DIfficulty should raise an error message if `0` or less as also if `11` or more. If the field is empty it should no error.

Year should raise an error if less than `1900` and if more than the current year (I added this validations, I don't think it makes sense to have a trick stablished at the future)

### DIFF
--- a/src/components/ui/customForm/TextInput.vue
+++ b/src/components/ui/customForm/TextInput.vue
@@ -10,6 +10,7 @@ import {
   FormLabel,
   FormMessage,
 } from '@/components/ui/form';
+import { InputHTMLAttributes } from 'vue';
 
 const props = defineProps<{
   title: string;
@@ -17,8 +18,8 @@ const props = defineProps<{
   formFieldName: string;
   placeholder?: string;
   class?: string;
-  inputMode?: string;
-  type?: string;
+  inputMode?: InputHTMLAttributes['inputmode'];
+  type?: InputHTMLAttributes['type'];
 }>();
 </script>
 
--- a/src/components/ui/customForm/TextInput.vue
+++ b/src/components/ui/customForm/TextInput.vue
@@ -20,6 +20,8 @@ const props = defineProps<{
   class?: string;
   inputMode?: InputHTMLAttributes['inputmode'];
   type?: InputHTMLAttributes['type'];
+  // This is used to pass values for the message translation interpolation
+  errorValues?: Record<string, string>;
 }>();
 </script>
 
@@ -36,7 +38,7 @@ const props = defineProps<{
       <FormDescription v-if="description">
         {{ description }}
       </FormDescription>
-      <FormMessage />
+      <FormMessage :values="props.errorValues" />
       <FormControl>
         <Input
           :type="type || 'text'"
--- a/src/components/ui/customForm/TextInput.vue
+++ b/src/components/ui/customForm/TextInput.vue
@@ -17,6 +17,8 @@ const props = defineProps<{
   formFieldName: string;
   placeholder?: string;
   class?: string;
+  inputMode?: string;
+  type?: string;
 }>();
 </script>
 
@@ -35,7 +37,12 @@ const props = defineProps<{
       </FormDescription>
       <FormMessage />
       <FormControl>
-        <Input type="text" :placeholder="placeholder" v-bind="componentField" />
+        <Input
+          :type="type || 'text'"
+          :placeholder="placeholder"
+          :inputMode="inputMode"
+          v-bind="componentField"
+        />
       </FormControl>
     </FormItem>
   </FormField>
--- a/src/components/ui/form/FormMessage.vue
+++ b/src/components/ui/form/FormMessage.vue
@@ -1,18 +1,10 @@
 <script lang="ts" setup>
 import { ErrorMessage } from 'vee-validate';
 import { toValue } from 'vue';
-import { useI18n } from 'vue-i18n';
-
-import messagesErrors from '@/i18n/error';
 
 import { useFormField } from './useFormField';
 
 const { name, formMessageId } = useFormField();
-
-const { t } = useI18n({
-  messages: messagesErrors,
-  scope: 'local',
-});
 </script>
 
 <template>
@@ -23,6 +15,6 @@ const { t } = useI18n({
     class="text-sm font-medium text-destructive"
     v-slot="{ message }"
   >
-    {{ message ? t(message) : '' }}
+    {{ message }}
   </ErrorMessage>
 </template>
--- a/src/components/ui/form/FormMessage.vue
+++ b/src/components/ui/form/FormMessage.vue
@@ -1,16 +1,15 @@
 <script lang="ts" setup>
 import { ErrorMessage } from 'vee-validate';
-import { toValue, computed } from 'vue';
+import { toValue } from 'vue';
 import { useI18n } from 'vue-i18n';
 
 import messagesErrors from '@/i18n/error';
-import { i18nMerge } from '@/i18n/i18nmerge';
 
 import { useFormField } from './useFormField';
 
 const { name, formMessageId } = useFormField();
 const { t } = useI18n({
-  messages: i18nMerge(messagesErrors),
+  messages: messagesErrors,
   scope: 'local',
 });
 
@@ -18,14 +17,6 @@ const props = defineProps<{
   // This is used to pass values for the message translation interpolation
   values?: Record<string, string>;
 }>();
-
-// Create a computed function to translate messages
-const translateMessage = computed(() => {
-  return (message: string) => {
-    if (!message) return '';
-    return t(message, props.values || {});
-  };
-});
 </script>
 
 <template>
@@ -36,6 +27,6 @@ const translateMessage = computed(() => {
     class="text-sm font-medium text-destructive"
     v-slot="{ message }"
   >
-    {{ message ? translateMessage(message) : '' }}
+    {{ message ? t(message, props.values || {}) : '' }}
   </ErrorMessage>
 </template>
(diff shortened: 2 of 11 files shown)


src/components/ui/customForm/TextInput.vue
```diff
@@ -17,6 +17,8 @@ const props = defineProps<{
   formFieldName: string;
   placeholder?: string;
   class?: string;
+  inputMode?: string;
```
JulianDietzel: What is `inputMode` used for? And is it possible to narrow the type `string` down to all the legal values (`"numeric" | "..."`)

src/components/ui/form/FormMessage.vue
```diff
@@ -23,6 +15,6 @@ const { t } = useI18n({
     class="text-sm font-medium text-destructive"
     v-slot="{ message }"
   >
-    {{ message ? t(message) : '' }}
+    {{ message }}
```
JulianDietzel: This appears to not update the translation immediately when the user selects a different language. It only updates once the page is reloaded. Since it works when reloading I don't think this is a big issue. If you have an idea on how to fix this easily it would be nice to have it fixed, if this will cause you troubles or be a bit of work just leave it as is.

I had a similar issue once and solved it by using Vues `watch`-method on `i18n.locales`. Maybe this is applicable here as well: https://github.com/bastislack/highline-freestyle/blob/ac70ce038fa9bf7a70a79e4993a3be365076643c/src/routes/tricks/TrickList.vue#L76

src/routes/tricks/NewTrickRoute.vue
```diff
@@ -28,20 +30,22 @@ const toast = useToast();
...
-    .int({ message: 'INPUT_NOT_INTEGER' })
-    .min(1, { message: 'INPUT_NUMBER_BELOW_MIN' })
-    .max(10, { message: 'INPUT_NUMBER_ABOVE_MAX' })
-    .optional(),
+  difficulty: z.union([
+    z
+      .number()
+      .int({ message: t('INPUT_NOT_INTEGER') })
+      .min(1, { message: t('INPUT_NUMBER_BELOW_MIN', { value: 1 }) })
+      .max(10, { message: t('INPUT_NUMBER_ABOVE_MAX', { value: 10 }) }),
```
JulianDietzel: I don't think we will get there any time soon. But since this is for the user to decide I would suggest we allow values of up to 20. No need to restrict the user more than necessary.

JulianDietzel (changes requested):
Thanks for the fix! It looks really good. I added some comments, the only one that I really want to see changed is the maximum difficulty the user can enter. The translation thing is such a minor issue that I don't think it needs to be fixed at this point if you don't really want to. If there is an easy quick fix though it would of course be nice

src/components/ui/customForm/TextInput.vue
Dosbodoke: It's for mobile keyboards, with inputMode numeric it will shown a numeric keyboard instead of the standard.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

Dosbodoke: I will type it :D

src/components/ui/form/FormMessage.vue
Dosbodoke: Yeah, the initial problem was that I couldn't pass parameters for interpolation.

Ideally, I would like the errorMessages to be imported only here and to have the following signature

`{{ message ? t(mesage, { ...messageParams }) : '' }}`

and, on the zod validation call it like `int({ message: 'INPUT_NOT_INTEGER', { value: 1 } })`

Not sure yet how I will solve this, let me take some time and iterations to figure it out.

src/components/ui/customForm/TextInput.vue
JulianDietzel: Thanks for the explanation, it makes sense now :+1:. Typing it sounds good to me. Looking at the docu you sent there shouldn't be too many option.

src/components/ui/form/FormMessage.vue
JulianDietzel: I don't know if I fully understand the issue but you seem to know a lot more about vee-validate so I'll just leave it up to you how or if you want to fix it.

Do you want to iterate on it in this PR or should we merge this and you fix it in another PR when / if you find the time for it?

Dosbodoke: Let me try to explain better the issue...

On the error translation files, for example `error.en.json` we have defined a space for interpolation

`"INPUT_NUMBER_BELOW_MIN": "Input must be no less than {value}",`

In order to this interpolation to work correctly we need to call it with `t("INPUT_NUMBER_BELOW_MIN", { value: 1})`

Where value is the exact same keyword to interpolate.

Initially I couldn't pass it down as a prop to `ErrorMessage` so I moved the translation to the zod definition

`.min(1, { message: t('INPUT_NUMBER_BELOW_MIN', { value: 1 }) })`

So we were not passing `INPUT_NUMBER_BELOW_MIN` as the message and translating on the child component. the message was already sent in it's final state.

But then it surges the error that you pointed out, the `t()` would run only when the user trigger the form validation (on submit or edit the field), and this not include changing the language.

I just made a [commit](https://github.com/Dosbodoke/highline-freestyle/commit/39a36c6b529f9e8f859460efe4b8ff2fd000df6a) to mitigate this issue on , let me know if you think it's a good pattern. This way we are passing down the translation key `INPUT_NUMBER_BELOW_MIN` and the necessary parameters for the interpolation.

src/components/ui/form/FormMessage.vue
```diff
@@ -1,18 +1,31 @@
...
 import messagesErrors from '@/i18n/error';
+import { i18nMerge } from '@/i18n/i18nmerge';
 
 import { useFormField } from './useFormField';
 
 const { name, formMessageId } = useFormField();
-
 const { t } = useI18n({
-  messages: messagesErrors,
+  messages: i18nMerge(messagesErrors),
```
WaldemarLehner: Is this intended? its been a long time since I looked at the code, but iirc this is used to merge i18n JSONs. "Merging" just 1 set should be identical to just passing it as-is

src/components/ui/form/FormMessage.vue
```diff
@@ -1,18 +1,31 @@
...
+  values?: Record<string, string>;
+}>();
+
+// Create a computed function to translate messages
+const translateMessage = computed(() => {
+  return (message: string) => {
+    if (!message) return '';
+    return t(message, props.values || {});
+  };
+});
```
WaldemarLehner: Am I right to assume that the computed is needed here to "re-generate" it on language change?

src/components/ui/form/FormMessage.vue
JulianDietzel: Thanks for the more detailed explanation. Form my understanding and talking with @WaldemarLehner  it looks good now :+1: .

We found two more things that were surprising to us, @WaldemarLehner commented on them.

src/components/ui/form/FormMessage.vue
Dosbodoke: It's a typo, functional, but yet a typo 😬

src/components/ui/form/FormMessage.vue
Dosbodoke: I'm more a React guy than a Vue one, really not sure how this ‘computed’ function works (I think it's like React ’useMemo’?), Claude gave me this code and I just certified it works

JulianDietzel: I fiddled around with it a little and I blieve Claude did a bit of an A"I" here. This does not have to be a computed function to work. The section can be simplified to:
```ts
function translateMessage (message: string): string {
  return (message) ? t(message, props.values || {}) : '';
}
```

I recognize it is getting quite nitpicky here but since this is a lot more readable, could you please swap out the section for the code I sent?

src/components/ui/form/FormMessage.vue
JulianDietzel: Makes sense, probably a leftover from how it was before. In that case please remove the `i18nMerge` function.

JulianDietzel:
One last thing and then I'll stop bugging you and merge this thing:

Since the difficulties can now go up to 20, the label on the field is incorrect in all languages. "1 to 10. Keep empty if unsure." should become "1 to 20. Keep empty if unsure.". The entries are the `question.difficulty` ones in the files in `/src/i18n/tricks/new/`.

Please change this. If you don't feel like it, it is totally fine, then just open a small issue for me to fix this later (so that it is not forgotten).

Dosbodoke:
> One last thing and then I'll stop bugging you and merge this thing:
> 
> Since the difficulties can now go up to 20, the label on the field is incorrect in all languages. "1 to 10. Keep empty if unsure." should become "1 to 20. Keep empty if unsure.". The entries are the `question.difficulty` ones in the files in `/src/i18n/tricks/new/`.
> 
> Please change this. If you don't feel like it, it is totally fine, then just open a small issue for me to fix this later (so that it is not forgotten).

Don't worry bro, I really missed it. Will push a fix soon :D

JulianDietzel (approved):
"soon", he said, pushing in less time than it took me to eat dinner :D. Great work! Looks really good to me :+1: . At last, I will now finally merge this :)

PR #376, opened 2025-03-24
Intial value to difficulty so it don't raise an wrong error

### LINKED ISSUE #373: Save personal trick - Difficulty can not be empty anymore
When creating a new personal trick the validation does not allow it anymore to leave the "Difficulty"-field empty. This should be possible in order for the difficulty to be undefined and therefore get the "Not determined" difficulty for tricks where the difficulty is not decided yet.

To solve this change the validation of the difficulty in such a way that the field can be empty.

Dosbodoke:
Fixes #373 

# The issue

After further exploration I found that the field was accepting an empty value, but it was necessary to type something and delete

The possible values for the fields are a `number` or an empty string `""` that enables it to be optional, but the initial value was `undefined`, making it fail on the zod validation

The data flow of typing and deleting was **undefined => value => ""**

# The fix

Add `difficulty: ''` to the useForm initialValues so it doesn't start as undefined and the validation works nicely.

# How to test

Open the `tricks/new` page, type a technical name and save it, it should work.

### DIFF
--- a/src/i18n/error/error.pt.json
+++ b/src/i18n/error/error.pt.json
@@ -3,7 +3,7 @@
   "home": "Início",
   "debug-info": "Informações de depuração",
 
-  "INPUT_REQUIRED": "Este campo é obrigatório",
+  "Required": "Este campo é obrigatório",
   "INPUT_NOT_NUMBER": "Por favor, insira um número",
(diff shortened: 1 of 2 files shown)


JulianDietzel:
Thanks for the explanation and the PR in general. Super clean fix, works as expected :+1:

Issue #355, opened 2025-01-29
Save personal tricks - Add menu to set variation of


JulianDietzel:
# Feature
In the menu to create new tricks the user should be able to specify 0 - n tricks that it is a variation of (a Double Sofa Roll is a variation of the regular Sofa Roll).

# Note
Consider that this will require a very similar menu as the one needed for #354. You probably want to reuse that one or (if this one is worked on first) create the component for this option as very general and reusable so that it can be used in the other PR as well.

Dosbodoke:
What do you suggest? A dropdown with some tricks as options? ( we may use a combobox for this https://www.shadcn-vue.com/docs/components/combobox.html )

Also, how do we will categorize tricks that can be used as these options?

JulianDietzel:
I have an idea for how to do this. I'll send a sketch of what I am thinking of the next couple of days.

As a sidenote: This issue is closely linked to #354, so we should consider the solution / design for these two issues as one.

JulianDietzel:
I was just about to present my idea for a design and then noticed that we have conflicting information in regards to the `variationOf` attribute. [In the initial planning, the wiki](https://github.com/bastislack/highline-freestyle/wiki/Backend-Overview#database-level) and the [currently used database schema](https://github.com/bastislack/highline-freestyle/blob/eea412676292cd9287d413b07f37c3fa16fea6fd/src/lib/database/schemas/Version1Schema.ts) `variationOf` is a list of tricks, while in this issue asks for 0 to 1 tricks for the `variationOf`. I would generally stick with the schema and the documentation but the version of this issue seems to make a lot more sense to me.

I've messaged @bastislack to resolve this inconsistency, so that we can move forward with this issue here.

In the meantime I will share my idea for adding multiple tricks in #354 as there we will for sure need a list of tricks.

JulianDietzel:
Talking to @bastislack, the backend, implementation and documentation are correct and this issue was incorrectly phrased by me. A trick can be a variation of multiple tricks. This is rare but possible. As an example he named the _Nerdflip_, which is a variation of both a _Boomerang_ and a _Rubik's Cube_.

I updated the description of this issue to the correct number of variation-of-tricks. As a result, the form fields of this issue and #354 will be identical and can be reused.

JulianDietzel:
Could you explain what you mean by this exactly @Dosbodoke ?

> Also, how do we will categorize tricks that can be used as these options?

JulianDietzel:
Addressed in #378

PR #378, opened 2025-03-29
feat: add "variation of" field to new trick page

JulianDietzel:
Addresses #355. It was a pain to implement but it works now.

Functional and translated into English, Spanish and French. The Portuguese translation is missing, I will open a PR for this shortly.

# Screenshot
Screen Shot 2025-03-28 at 19 19 31

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "dexie": "^3.2.4",
         "lucide-vue-next": "^0.294.0",
         "radix-vue": "^1.9.10",
+        "reka-ui": "^2.1.0",
         "tailwind-merge": "^2.0.0",
         "tailwindcss-animate": "^1.0.7",
         "vee-validate": "^4.15.0",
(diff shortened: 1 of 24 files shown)


JulianDietzel:
Related translation issue: #380

bastislack:
Awesome! It looks great :)
What were the most painful parts about the implementation?

JulianDietzel:
Thanks :), glad you like it.

The main issue for me was understanding how the form data is handled. There is an interplay between vue, vee-validate and zod in the form and I had little knowledge of zod and close to no knowledge of vee-valide up to this point. All those libraries do play together but have their own opinion how data should be handled and passed around. All this was not helped by the fact that some of the components I used came from shadcn, yet another source, with again its own opinions on the matter.

Then on top of that I had to figure out how exactly @WaldemarLehner dealt with all of this almost two years ago in his implementation to adapt to it. (And I do recall him having a bit of a fight with shadcn as well back then)

Finally add the standard humanness of missing an important line here or making a reasonable but false assumption there and the frustration was complete.

Do I have the phrase "I hate frameworks" in my DuckDuckGo search queries now? Perhaps. But man was it a relief when it finally clicked and worked yesterday evening :D. And I think I understand the system well enough now that new form components should be a lot easier to implement.

PR #383, opened 2025-04-01
#380 Portuguese translation for MultiTrickSelectInternal.vue

### LINKED ISSUE #380: Translate MultiTrickSelect component to Portuguese
#378 introduced the new MultiTrickSelect component (`src/components/ui/customForm/MultiTrickSelect` and `src/components/ui/customForm/MultiTrickSelectInternal`). As you can see, it is comprised of two files / components. The relevant one for translation is the "Interal" one.

I designed it in English and translated it into Spanish and French. The translation into Portuguese is still missing.

This should be a fairly small issue as only two fields need to be translated.

Dosbodoke:
Resolves #380


JulianDietzel:
Thanks for the quick translation <3

PR #386, opened 2025-04-13
Parse unknown difficulty levels

bastislack:
In the trick list some tricks have '?' as difficultyLevel. This now gets converted to 'tbd' and stored in the trick objects.

### DIFF
--- a/scripts/fetchFromSheet.ts
+++ b/scripts/fetchFromSheet.ts
@@ -196,7 +196,7 @@ const trickObjects = trickRawObjects.map((csvTrick) =>
       .map((e) => e.trim())
       .filter(Boolean),
     videos: videoLookup['trick-' + csvTrick.id],
-    difficultyLevel: Number(csvTrick.difficultyLevel),
+    difficultyLevel: csvTrick.level === '?' ? 'tbd' : Number(csvTrick.level),
     showInSearchQueries: csvTrick.showInSearchQueries === 'TRUE',
     dateAddedEpoch: new Date(csvTrick.dateAddedIso8601).getTime(),
     yearEstablished: csvTrick.yearEstablished ? Number(csvTrick.yearEstablished) : undefined,
--- a/scripts/fetchFromSheet.ts
+++ b/scripts/fetchFromSheet.ts
@@ -196,7 +196,7 @@ const trickObjects = trickRawObjects.map((csvTrick) =>
       .map((e) => e.trim())
       .filter(Boolean),
     videos: videoLookup['trick-' + csvTrick.id],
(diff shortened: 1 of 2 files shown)


bastislack:
Closes #266.

scripts/fetchFromSheet.ts
```diff
@@ -196,7 +196,7 @@ const trickObjects = trickRawObjects.map((csvTrick) =>
       .map((e) => e.trim())
       .filter(Boolean),
     videos: videoLookup['trick-' + csvTrick.id],
-    difficultyLevel: Number(csvTrick.difficultyLevel),
+    difficultyLevel: csvTrick.level === '?' ? 'tbd' : Number(csvTrick.level),
```
JulianDietzel: I'd suggest instead of starting to mix numbers and strings, we solve this by having the "To be Determined" state be represented by `undefined`. This would make handling the the case easier (Consider for example that the `trick.difficultyLevel ?? ...` syntax of javascript could be used with this way of doing it). I'd therefore suggest changing `'tbd'` to `undefined` in this line.

src/lib/database/schemas/Version1Schema.ts
```diff
@@ -53,7 +53,7 @@ export const DbTricksTableZod = z.object({
   alias: z.string().nonempty().optional(),
   startPosition: DbPositionZod,
   endPosition: DbPositionZod,
-  difficultyLevel: z.number().int().min(0).optional(),
+  difficultyLevel: z.union([z.number().int().min(0), z.literal("tbd")]).optional(),
```
JulianDietzel: As a follow up to the previous comment:

If we go with the option of `undefined` = To Be Determined, then we don't need any changes here at all, since the `.optional()` in the end already allows us to pass in `undefined`.

JulianDietzel (commented):
Firstly thanks for the PR <3.

I'd suggest using `undefined` instead of the string literal "tbd". See the other comments for more context.

Something I have not addressed in the other comments is that in the few cases I have come across having to use the To Be Determined state when writing the frontent, I have made the assumption that `undefined` = To Be Determined. So some components already expect this logic (for example the difficulty label on the Trick Detail page). If we want to move away from this logic / assumption / way of doing it we need to consider that more components need to be updated.

But before we get into that, let me know what you think of my suggestion and if you still feel like we should go with the mix of the string literal and number type.

scripts/fetchFromSheet.ts
bastislack: Totally makes sense :) Just changed it.

JulianDietzel:
There are still other errors with the import script so I wasn't able to test if this works correctly in all cases. I'll therefore merge this PR and then address the other issues in a followup PR.

PR #403, opened 2025-09-11
feat: initially load current state of official tricks

JulianDietzel:
Importer via the fetchFromSheets.ts script.


JulianDietzel:
Siiick. Getting a whole lot closer. Only one left :D

bastislack (approved):

JulianDietzel:
The one trick yml failing might point to a more general issue. I don't know if this is the issue yet but the description of Trick 243 contains a colon which is also used in yml syntax. Maybe thats the issue?

bastislack:
:)

PR #412, opened 2025-09-28
Loop youtube videos

### LINKED ISSUE #398: Loop YouTube videos
Currently YouTube videos with start and end time are not looped by default

bastislack:
YouTube videos are now looping automatically.

I was thinking about removing the `Duration` component, as now I don't see any value in it, what do you think?

Closes #398


JulianDietzel:
Awesome! It works again. Thanks :)

On removing the `Duration` component: I don't think we want to remove the component completely. We might consider not showing it on YouTube videos as the are now "cut" to the length specified as the duration. But for Instagram videos or video links which are neither YouTube nor Instagram videos it is still very handy to display the start and end times as numbers (see my [comment under the Instagram PR](https://github.com/bastislack/highline-freestyle/pull/414#discussion_r2395685417) for a little more info)

As for if we want to keep the duration displayed as a start and end time on YouTube videos: I don't care all that much either way. If you have a clear preference one way or the other we can just do it that way. I think the only but real benefit to removing it would be to remove visual clutter.

.eslintrc.cjs
```diff
@@ -39,5 +39,8 @@ module.exports = {
     // vue/multi-word-component-names is disabled as it doesn't make much sense in our opinion.
     // "Navbar" for example is descriptive as-is and doesnt require another word.
     "vue/multi-word-component-names": "off"
+  },
+  globals: {
+    YT: "readonly"
```
JulianDietzel, line 44: Don't know 100% what is going on here. But if you are, then it's fine.

JulianDietzel (approved):
Awesome! It works again. Thanks :)

On removing the `Duration` component: I don't think we want to remove the component completely. We might consider not showing it on YouTube videos as the are now "cut" to the length specified as the duration. But for Instagram videos or video links which are neither YouTube nor Instagram videos it is still very handy to display the start and end times as numbers (see my [comment under the Instagram PR](https://github.com/bastislack/highline-freestyle/pull/414#discussion_r2395685417) for a little more info)

As for if we want to keep the duration displayed as a start and end time on YouTube videos: I don't care all that much either way. If you have a clear preference one way or the other we can just do it that way. I think the only but real benefit to removing it would be to remove visual clutter.

.eslintrc.cjs
bastislack: The problem was that eslint kept telling me that "YT is not defined", even though it clearly worked and I was getting all the autocompletion I could hope for. I tried many things and this was the only thing that fixed it ;)

bastislack:
Since we don't need the Duration on the IG videos (see my comment there) and here it basically only shows redundant information (and currently there are no other video-platforms we support), I'm tempted to remove it completely (and when we need it in the future, it's as easy as reverting that commit). Let me know what you think!

PR #414, opened 2025-09-28
fix: Use correct IG embed url

### LINKED ISSUE #399: Fix instagram videos
Instagram videos currently don't work

bastislack:
closes #399

### DIFF
--- a/src/components/video/platforms/InstagramPlayer.vue
+++ b/src/components/video/platforms/InstagramPlayer.vue
@@ -1,13 +1,24 @@
 <script setup lang="ts">
 import { isEmbedAllowed } from '@/util/trackingPreferences';
 import EmbedPrompt from '../EmbedPrompt.vue';
-import Duration from '../Duration.vue';
+import { computed } from 'vue';
 
-defineProps<{
+const props = defineProps<{
   url: string;
   startTime?: number;
   endTime?: number;
 }>();
+
+const embedUrl = computed(() => {
+  let baseUrl = props.url.trim();
+
+  // Guarantee it ends with a slash before appending "embed"
+  if (baseUrl.endsWith('/')) {
+    return baseUrl + 'embed';
+  } else {
+    return baseUrl + '/embed';
+  }
+});
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


src/components/video/platforms/InstagramPlayer.vue
```diff
@@ -16,17 +27,11 @@ defineProps<{
...
         scrolling="no"
         title="video"
       ></iframe>
     </div>
-    <Duration
-      v-if="startTime || endTime"
-      :start="startTime"
-      :end="endTime"
-      class="mx-auto mt-2 text-muted-foreground"
-    />
```
JulianDietzel, line 30: I think that it is nice to keep the duration component. Without it the user does not know when the tricks in the video starts and ends. Instagram videos are usually shorter so this isn't necessary often but there are videos with muliple tricks in it and then it could come in handy (if no start or end time is specified the duration isn't shown anyways).

With YouTube videos the argument could be made that we can simply only show the duration specified but we don't seem to be able to do this with instagram, so no way around showing the duration in numbers to the user, I think.

JulianDietzel (changes requested):
A little sad that we (or at least I) can't display the videos directly in the app anymore but I'm glad you got the previews to to render again. This is a lot nicer and it works on my machine too. To see the videos it redirects me to another tab though  but I guess this is intended.

Apart from my comment on the `Duration` I am happy with this :)

src/components/video/platforms/InstagramPlayer.vue
bastislack: I removed it under the IG videos because I only added Start + End time to YT videos, so it will never be shown here. Since IG doesn't support the looping for a certain timeframe (and it is actually not easy to see the timestamps that are interesting in the video), I never bothered to try to even track these infos.

bastislack: Let me know if this is ok for you!

bastislack:
> A little sad that we (or at least I) can't display the videos directly in the app anymore 

Yes this is really annoying about IG... I hope we can find a better solution for this at some point, but for now I think it's good enough that users can watch the videos somehow.

PR #428, opened 2025-11-02
Show variations

bastislack:
Tricks that have variations now have a dropdown button that expands a subgrid with their variations.
There have also been some layout changes to the cards in order to clean everything up.


src/assets/tricks/1.yaml
bastislack, line 1: Not sure why these files show up in this PR since they are deleted on the current 'rewrite' already...

src/components/stickable/list/StickableCard.vue
```diff
@@ -0,0 +1,71 @@
...
+
+const props = defineProps<{
+  to: string; // when present, render as RouterLink
+  class: string;
+}>();
+
+const textElement = ref<HTMLElement>();
+const isOverflowing = ref(false);
+
+const checkOverflow = () => {
```
bastislack: Please let me know if this is a very hacky solution.. It was created by my fellow AI-companion, but since I don't have a smarter solution I just used it..

src/components/stickable/list/VariationsCollapsibleWrapper.vue
bastislack, line 1: Also all the logic in this file was mainly created using AI. I spent a lot of time on this but this was the best approach in terms of how it works in the app. Let me know if this code can be simplified or written in a cleaner way.

src/components/ui/popover/Popover.vue
bastislack, line 1: Added via shadcn-vue

src/components/ui/popover/PopoverContent.vue
bastislack, line 1: Added via shadcn-vue

src/components/ui/popover/PopoverTrigger.vue
bastislack, line 1: Added via shadcn-vue

src/components/ui/popover/index.ts
bastislack, line 1: Added via shadcn-vue

src/services/searchAndFilterTricks.ts
bastislack, line 1: I changed the logic so now the favorites are shown at the top AND where-ever they were before. This seems more intuitive than when they disappear from their original place.

src/components/stickable/list/VariationsCollapsibleWrapper.vue
```diff
@@ -0,0 +1,137 @@
...
+      pendingOpenId.value = null;
+      openCollapsibleId.value = idToOpen;
+    }
+  }
+}
+
+watch(isOpen, async (open) => {
+  if (open) {
+    isAnimating.value = true;
+    await nextTick();
```
JulianDietzel: Those two lines are the same in both the if an else branch. They should be moved above the if statement.

src/components/stickable/list/FavoriteMark.vue
JulianDietzel, line 1: Don't know if this really needs to be an extra component. It seems like it is only used it one place. Keep it like this if you want but I would move the code into the also still really small `CardDecoration` component.

src/components/stickable/list/StickableCard.vue
```diff
@@ -0,0 +1,71 @@
+<script setup lang="ts">
+import { RouterLink } from 'vue-router';
+import { ref, onMounted, onBeforeUnmount, nextTick } from 'vue';
+
+const props = defineProps<{
+  to: string; // when present, render as RouterLink
```
JulianDietzel, line 8: Is this comment true? Seems to me like it would always be a RouterLink as this is the base element of the template. But perhaps there is some funky behavior of the RouterLink component when the `to` property is missing?

src/components/stickable/list/StickableCard.vue
JulianDietzel: Nitpick: Could perhaps be renamed to `updateIsOverflowing`, making the side effect of modifying the `isOverflowing` clear at first sight.

JulianDietzel: If the solution does not cause flickering when the element is removed and readded I'm happy with some hackyness for visual stuff :D. A creative solution for sure. I'd add a short comment to the code explaining that the function is in fact a bit hacky and explain what is happening in two or three sentences.

src/i18n/list/list.en.json
```diff
@@ -6,10 +6,19 @@
...
+        "title": "Personal",
+        "description": "This is a personal trick that you added to your trick-list. No-one else will be able to see this trick in their trick list."
+      },
+      "archived": {
+        "title": "Archived",
+        "description": "This trick used to be part of the official trick-list but got removed at some point."
+      },
+      "new": {
+        "title": "New",
+        "description": "This trick has been recently added to the official trick-list. Go out and try to land it!"
```
JulianDietzel: Nitpick: I'd write "trick list" instead of "trick-list". Doesn't matter much though, leave as is if you want to

src/components/stickable/list/StickableOverviewCard.vue
JulianDietzel, line 1: Either this component (/file) or `StickableCard.vue` needs a new name. It is not clear what the difference between them is. I haven't fully managed to understand all the things that this file does so I don't have solid recommendation yet. Maybe this could be the `StickableCardExtendable.vue`? I'm open for better ideas.

src/components/stickable/list/CardDecoration.vue
```diff
@@ -0,0 +1,23 @@
...
+<template>
+  <slot
+    name="decoration"
+    :isFavorite="props.isFavorite"
+    :isNew="props.isNew"
+    :status="props.status"
+  >
+    <FavoriteMark :show="props.isFavorite" />
+    <StatusIcon :isNew="props.isNew" :status="props.status" />
+  </slot>
```
JulianDietzel: Why was a slot used here? I mean yes, I guess the default content could be overwritten by inserting something into the slot but that would, in my understanding, lead to the inserted content being displayed as if the CardDecoration component was not there at all. Is there a reason why this is a slot?

If there is no good reason then this component breaks down into almost nothing and could potentially be fully removed, simply replacing every reference to it with lines 20 and 21.

I could very well be missing something though.

src/components/stickable/list/StickableCard.vue
```diff
@@ -0,0 +1,71 @@
...
+});
+</script>
+
+<template>
+  <RouterLink
+    :to="to"
+    class="p-2 rounded-sm border aspect-[7/5] flex text-center relative"
+    :class="props.class"
+  >
+    <slot name="decoration" />
```
JulianDietzel: I'm doing this review purely on GitHub right now so I don't have all my dev tools at hand. But from what I have seen this slot is always populated with the same `CardDecoration` component. If that is the case this doesn't need to be slot and can simply always contain the `CardDecoration`.
Optionally the slot could be kept with the `CardDecoration` as the default content. Then in 90% of all cases it is cleaner / shorter to use this (`StickableCard`) component while still allowing the decoration to be replaced. Though honestly, if it is not needed then I wouldn't even do that.

JulianDietzel: Since I proposed potentially dissolving the `CardDecoration` component, its content could maybe just be inserted here.

src/components/stickable/list/StickableOverviewCard.vue
```diff
@@ -19,10 +12,12 @@ const props = defineProps<{
   stickFrequency?: number;
   isFavorite: boolean;
   isNew: boolean;
+  variations: SearchItem[];
+  showVariations: boolean;
 }>();
 
-const borderClass = computed<string>(() => {
-  let stickFrequency = Math.max(0, Math.min(props.stickFrequency ?? 0, 7));
+function getBorderClass(stickFrequency?: number): string {
```
JulianDietzel: Nitpick: Not a big fan of prefixing functions with `get`. Seems useless to me 90% of the time while making the name longer. But not a real issue either :D.

src/components/stickable/list/StickableOverviewCard.vue
```diff
@@ -46,51 +41,53 @@ const fillClass = computed<string>(() => {
...
-    <div v-if="isFavorite">
-      <Icon icon="ic:round-star" class="absolute top-1 right-1 h-5 w-5 z-10" />
-      <div class="absolute top-0 right-0 w-7 h-7 z-0 blur-md rounded-full" :class="fillClass"></div>
-      <div class="absolute top-0 right-0 w-7 h-7 z-0 blur-md rounded-full" :class="fillClass"></div>
-      <div class="absolute top-0 right-0 w-6 h-6 z-0 blur-sm rounded-full" :class="fillClass"></div>
-      <div class="absolute top-0 right-0 w-6 h-6 z-0 blur-sm rounded-full" :class="fillClass"></div>
-    </div>
-
-    <div class="flex-grow flex flex-col justify-around w-full">{{ title }}</div>
+    <StickableCard :to="props.linkToDetails" :class="getComputedClass(props.stickFrequency)">
```
JulianDietzel: While this approach is ok, I believe it would be cleaner if the `StickableCard` would set its classes (and therefore its color and border) itself based on the stick frequency. I would propose for all the code of `getComputedClass` and all the functions it calls to be moved into the `StickableCard` component and adding a `stickFrequency` prop to the it. Then let the component do the computation itself.

Not strictly necessary but I think this would make the code cleaner and more readable. Also the StickableCard could be reused elsewhere and maintain consistent colors as the don't need to be inserted from the outside via the class attribute.

src/components/stickable/list/StickableOverviewCard.vue
```diff
@@ -46,51 +41,53 @@ const fillClass = computed<string>(() => {
...
-      <Badge
-        v-if="isNew"
-        class="px-2 bg-secondary-950 text-secondary-50 hover:bg-secondary-950"
-        variant="secondary"
+    <template #variations>
+      <StickableCard
+        v-for="variation in variations"
+        :key="`${variation.primaryKey[1]}:${variation.primaryKey[0]}`"
+        :to="variationLinkToDetails(variation.primaryKey)"
+        :class="getComputedClass(variation.stickFrequency)"
```
JulianDietzel: See above comment.

src/components/stickable/list/StickableOverviewCard.vue
```diff
@@ -46,51 +41,53 @@ const fillClass = computed<string>(() => {
...
+            :isNew="variation.isNew"
+            :status="variation.primaryKey[1]"
+          ></CardDecoration>
+        </template>
+        {{ variation.name }}
+      </StickableCard>
+    </template>
+  </VariationsCollapsibleWrapper>
+
+  <StickableCard v-else :to="props.linkToDetails" :class="getComputedClass(props.stickFrequency)">
```
JulianDietzel: See above comment.

src/components/stickable/list/VariationsCollapsibleWrapper.vue
```diff
@@ -0,0 +1,137 @@
...
+  ro = null;
+});
+</script>
+
+<template>
+  <!-- Use a wrapper div instead of making Collapsible use 'contents' -->
+  <div class="contents">
+    <Collapsible v-model:open="isOpen" class="contents">
+      <div class="relative">
+        <!-- Parent -->
```
JulianDietzel: Is the following slot _really_ a parent? Looks like a child to me.

src/components/stickable/list/VariationsCollapsibleWrapper.vue
JulianDietzel: I'm not all to happy with AI code noone understands but I don't have the capacities right now to deep dive into all that is going on here. So if it works I guess hope it is fine.

src/components/stickable/list/OpenCollapsibleId.ts
JulianDietzel, line 1: Why is this inside a new "top" level `key` directory instead of either directly in the `src/components/stickable/list/` or `src/util` directory?

src/routes/tricks/TrickList.vue
```diff
@@ -93,6 +101,37 @@ watch(
...
+    const newVariationsMap = new Map<string, SearchItem[]>();
+    searchResult.value?.forEach((section) => {
+      section.items.forEach((item) => {
+        const variations = getVariationsForTrick(
+          allTricks,
+          item.primaryKey[0],
+          item.primaryKey[1],
+          searchParameters.value.includedStatuses
+        );
+        if (variations.length > 0) {
```
JulianDietzel: Consider an Early Return for shallower nesting / indentation. (`if (variations.length) == 0) return`)

JulianDietzel (changes requested):
Thanks for all the work. Haven't run the code but only looked at it so far. Here is the stuff I noticed.

For next time it would be easier for me to have a smaller MR to review. I couldn't really do it justice as I just don't have the capacities to check this much code after a day of reading a lot of code already. This kind of feature would have resulted in a large MR no matter what you do but I feel like the removal of the .yml files and the changes to the handling of the favorites could have been in two extra MRs.

Thanks again though!

src/components/stickable/list/StickableCard.vue
bastislack: This was probably created by AI and I missed it

src/components/stickable/list/StickableCard.vue
bastislack: Added a comment to explain the hacky solution :D

src/components/stickable/list/VariationsCollapsibleWrapper.vue
bastislack: Not too happy with this code either, but it seems to work and without it this feature would have taken forever :D

src/components/stickable/list/StickableOverviewCard.vue
bastislack: I renamed `StickableOverviewCard` to `StickableSearchResult` because this component is what appears on the top-level in the trick list and then checks whether there are any variations, then conditionally shows these. Also the colors for the stickFrequency are applied here and passed to the underlying `StickableCard`.

src/components/stickable/list/CardDecoration.vue
bastislack: I initially thought it's nice to have it be overridable, but since we're not using it I removed it again. I'd still keep the Decoration component for now because I feel there is gonna be more things going there at some point (e.g. the number that shows the difficulty of the variations).

src/components/stickable/list/StickableCard.vue
bastislack: I think one reason, why I kept the Decoration in this component (filling its slot), is because it saves me one pass of the props (isFavorite, etc...). But I guess this is not a reason enough to make this design decision. I moved the Decoration directly into the StickableCard now

src/components/stickable/list/StickableOverviewCard.vue
bastislack: True it does look much better now

src/components/stickable/list/VariationsCollapsibleWrapper.vue
bastislack: I removed the comment, since I think it's not really necessary. But in terms of the collapsible that is the 'parent card' from which all the variations are then shown.

JulianDietzel (approved):
Looks good now :+1:

JulianDietzel:
I just ran the code and noticed that on my machine sometimes the collapsible menu flickers, which then causes the texts of the cards two rows down to flicker (at a different rythm). If the wrong row flickers, the heading below it will jiggle up and down at an irregular interval.

Also a lot more tricks are now hidden in submenus than I expected. Ill play with it a bit more and we can talk about this in the next meeting.

Issue #400, opened 2025-09-11
First time opening the app there are no tricks


bastislack:
Load tricks when app is opened for the first time

JulianDietzel:
Never mind, testing this can be a bit tricky, the issue still persists despite the attempted fix in #409

JulianDietzel:
This now seems to be fixed with #429 by strongly reducing the synchronization times of tricks to the database. But I am not 100% sure if this actually was the issue, it just was the only thing I could reliably reproduce (very long loading times).

We'll see if this is the final fix. If not, just reopen the issue (or message me to reopen it if you don't have the permissions to do so)

PR #429, opened 2025-11-07
feat: improve loading time for syncing tricks

JulianDietzel:
Adresses #400 (hopefully).

I assume the issue persisted because loading could take seconds and therefore the impression of the tricks not loading at all could arise. Who knows if this was the issue but it is the only thing I could reproduce.

As a result I improved the loading performance when syncing the tricks from an average of 1957ms (std. dev. 394ms, 5 samples) down to 327ms (std. dev. 40ms, 5 samples) when running a production build (`npm run build && npm run preview`) (though the numbers don't differ significantly when running in development mode (`npm run host`). A performance improvement of around 8x.

This was archieved by fully removing the functionality to unarchive tricks. The function for this was implemented in an inefficient way where it would cause reads (and potential writes) for every trick in the Index DB, whether or not it had to be unarchived. This removal alone was responsible for the performance improvement.

But Julian, I hear you say, you can't simply improve performance by throwing out functionality at the core of the app. That's what I thought too. At first. But as I pondered what unarchiving actually does and the edge cases it has I came to the conclusion that removing it was the best step forward:

Unarchiving did the following: If there was a trick with an id `A` and a status `archived` in the Index DB and there was a trick with the same id `A` in the trick yml files to be added to the database, the function would detect that and
1. Create a new Trick with id `A` and status `official` in the database
2. Add all the information of the trick in the yml files to the entry in the database
3. Point the metadata of the original archived trick in the database to the new entry
4. Finally delete the old entry

This way if a trick was removed from the official tricks but was added back again, the metadata of the user could be preserved for that trick. That is, _if_, and only if, the user didn't choose to convert the archived trick into a userDefined trick yet. If the user does that there is no way to safely unarchive it again as official tricks and userDefined tricks can have the same id whilst being completely independant and different tricks.

So unarchiving was used in the rare case that a trick would be removed from the official trick and added back again. That shouldn't be a common occurrence. And then in addition to that it doesn't work in all cases. And yet on top of that, it does not cause any issues if this behavior does not exists. If unarchiving is not in the code and an official trick is reintroduced, then it will simply be in the users trick list twice. Once as official, once archived. Not great but it won't be causing any bugs either, plus, it is actually fairly understandable to the user. They see two versions of the trick, one with an "archived" Badge and old data, one that looks like an official trick but with new data. For something that shouldn't happen in the first place this seems reasonable enough.

### DIFF
--- a/src/lib/database/official/syncTricks.ts
+++ b/src/lib/database/official/syncTricks.ts
@@ -152,32 +152,6 @@ async function handleArchiving(archivedTricks: z.infer<typeof DbTricksTableZod>[
   return failedTrickMigrations === 0;
 }
 
-/**
- * Although the odds of a trick that was archived being reintroduced are very, it might still be possible.
- * This will
- * - find metadata belonging to an archived trick and move it to the official trick
- * - move any references to the archived trick back to the official trick
- * - delete the archived trick (creation will happen later)
- */
-async function handleUnarchivingTricks(tricks: z.infer<typeof DbTricksTableZod>[]) {
-  return await Promise.allSettled(
-    tricks.map(async (e) => {
-      await moveTrick(e.id, 'archived', 'official');
-      // At this point references are handled. Now we just need to move the actual Trick over.
-      // This is done using a PUT and DELETE (because again, you cannot modify the Primary Key with Dexie / IndexedDB)
-      const newTrickObject = DbTricksTableZod.parse({
-        ...e,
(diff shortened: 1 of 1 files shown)


bastislack:
Yes, amazing performance improvement!!
I really think that is an edge case that will rarely occur, so this is great :)

bastislack (approved):

